PR #1, opened 2023-08-31
add convinience functions

kissiel:
The functions added here were previously put in (virtually) every jenkins job. Now those functions are stand-alone scripts, and because the scripts have the same filenames as the function from the jobs, they serve as a drop-in replacement.

### DIFF
--- a/scriptlets/README.md
+++ b/scriptlets/README.md
@@ -1,0 +1,15 @@
+# Convinience functions for Jenkins jobs
+
+## description
+
+This directory contains convinience functions that are used in almost all of Certification Team's Jenkins.
+
+## testing
+
+The unit tests are written in [BATS (Bash automated Testing System)](https://github.com/bats-core/bats-core). To run the tests run:
+
+```bash
+
+cd tests
+bats .
+```
--- a/scriptlets/_put
+++ b/scriptlets/_put
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+#
+# Copyright 2023 Canonical Ltd.
+#
+# This script provides a mechanism for transferring a file to a remote device via SCP.
+# It uses the `set-ssh-options` script to set SSH options and then executes the SCP command.
+# This script is intended to be a drop-in replacement for inline convenience
+# functions previously defined in each of the Jenkins jobs.
+#
+# Dependencies:
+#     This script depends on the `set-ssh-options` script to set SSH options.
+#     DEVICE_IP and DEVICE_USER must be set in the environment.
+#
+# Usage:
+#     `_put <local_file> <remote_file>`
+#     Or source it in another script and call
+#     `put_function <local_file> <remote_file>`
+
+put_function() {
+  source "$(dirname "$BASH_SOURCE")/set-ssh-options"
+  scp $SSH_OPTS $1 $DEVICE_IP:$2
+}
+
+# Check if the script is being sourced or executed
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
+  put_function "$@"
+fi
--- a/scriptlets/_run_retry
+++ b/scriptlets/_run_retry
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+# Copyright 2023 Canonical Ltd.
+#
+# This script provides a retry mechanism for running a command.
+# It uses the `_run` function to execute the command and checks the return
+# code. If the command fails, it will retry up to 20 times,
+# with a 30-second sleep between each attempt.
+# This script is intended to be a drop-in replacement for inline convinience
+# functions previously defined in each of the Jenkins jobs.
+#
+# Usage:
+#     `_retry_run <command> [command_args ...]`
+#     Or source it in another script and call
+#     `retry_run <command> [command_args ...]`
+
+retry_run() {
+  loopcnt=0
+  _run "$@"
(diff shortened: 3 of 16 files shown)


scriptlets/README.md
```diff
@@ -0,0 +1,15 @@
+# Convinience functions for Jenkins jobs
+
+## Description
+
+This directory contains convinience functions that are used in almost all of Certification Team's Jenkins.
```
pieqq, line 5: I think it's "convenience".

scriptlets/_put
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+#
+# Copyright 2023 Canonical Ltd.
+#
+# This script provides a mechanism for transferring a file to a remote device via SCP.
+# It uses the `set-ssh-options` script to set SSH options and then executes the SCP command.
+# This script is intended to be a drop-in replacement for inline convenience
+# functions previously defined in each of the Jenkins jobs.
```
pieqq, line 8: I feel like this should be part of the commit body and not of the script itself.

scriptlets/_run_retry
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+# Copyright 2023 Canonical Ltd.
+#
+# This script provides a retry mechanism for running a command.
+# It uses the `_run` function to execute the command and checks the return
+# code. If the command fails, it will retry up to 20 times,
+# with a 30-second sleep between each attempt.
+# This script is intended to be a drop-in replacement for inline convinience
+# functions previously defined in each of the Jenkins jobs.
```
pieqq, line 10: Same comment as above. It should live in the commit body instead.

scriptlets/wait_for_snap_complete
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+
+# Copyright 2023 Canonical Ltd.
+#
+# This script waits for all snap operations on a remote device to complete.
+# It uses SSH to query the snap changes on the device and checks for any ongoing operations.
+# The script will retry up to 40 times, with a 30-second sleep between each attempt.
+# This script is intended to be a drop-in replacement for inline convenience
+# functions previously defined in each of the Jenkins jobs.
```
pieqq, line 9: same same.

scriptlets/wait_for_ssh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+#
+# Copyright 2023 Canonical Ltd.
+#
+# This script provides a mechanism for waiting until SSH is available on a remote device.
+# It uses the `set-ssh-options` script to set SSH options and then repeatedly attempts to
+# connect to the device via SSH until successful or until a timeout is reached.
+# This script is intended to be a drop-in replacement for inline convenience
+# functions previously defined in each of the Jenkins jobs.
```
pieqq, line 9: One, two, you know what to do!

pieqq (changes requested):
Whoa!

And QA and I were discussing about unit testing in Bash literally 4 hours ago! I didn't know about BATS, it's pretty neat!

Nicely cleaned up functions, unit tested, and documented. It's +1000 from me!

Just a little comment regarding some comments in the functions that, in my opinion, should live in the commit body instead of the function itself, but feel free to ignore.

scriptlets/wait_for_snap_complete
```diff
@@ -0,0 +1,50 @@
...
+        loopcnt=$((loopcnt+1))
+        changes="$(ssh $SSH_OPTS $DEVICE_IP 'sudo snap changes')"
+        echo Checking for all snap operations to complete...
+        echo $changes
+        if [ -n "$changes" ] &&
+            [[ "$changes" != *"Doing"* ]] &&
+            [ "$(ssh $SSH_OPTS $DEVICE_IP echo true)" = "true" ]; then
+            done="true"
+            break
+        else
```
plars, line 35: minor nitpick on things like this, where you can just as well skip the else and just do
```
...
    fi
    sleep 30
done
```

scriptlets/wait_for_ssh
```diff
@@ -0,0 +1,38 @@
...
+#     `wait_for_ssh_function`
+
+wait_for_ssh_function() {
+    source "$(dirname "$BASH_SOURCE")/set-ssh-options"
+    loopcnt=0
+    until timeout 10 ssh $SSH_OPTS $DEVICE_IP /bin/true
+    do
+        echo "Testing to see if system is back up"
+        loopcnt=$((loopcnt+1))
+        if [ $loopcnt -gt 20 ]; then
```
plars, line 26: There have been a few times in the past where we needed to adjust this loopcnt (or optionally the sleep below... either has the same effect of giving more time). It might be good to default it to 20, but allow for an optional argument here

plars (changes requested):
A couple of very minor suggestions. Other than that, what do you think about maybe including some convenient file to source to pull it in to the path?
Maybe something like:
`. hwcert-jenkins-tools/scriptlets/import-all` which does something like:
```
SCRIPT=$(readlink -f $0)
export PATH=$PATH:$(dirname $SCRIPT)
```

Of course, another alternative is for everything to just manually add them to the path, but it might be worth mentioning in the README?
Looks good though, thanks for doing this!

plars (approved):
Looks great, this is nice! +1

PR #3, opened 2023-10-03
add script that provisions Checkbox from source

kissiel:
This enables CI/CD pipes to use checkbox from source without having to explicitly run all of the commands contained here

Resolves CHECKBOX-1063

### DIFF
--- a/snap-update-tools/provision_checkbox.sh
+++ b/snap-update-tools/provision_checkbox.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+
+# This script is used to provision the controller node with the necessary dependencies
+# and to install the latest version of Checkbox from github.
+
+# the `set`s below is used to make sure the script fails if any of the commands fail
+# and to print the commands as they are executed
+set -x
+set -e
+
+# setup what's needed for pip
+sudo apt-get update
+sudo apt-get install python3-pip python3-dev python3-virtualenv build-essential -y
+
+# clone the Checkbox repo from github if it doesn't exist, otherwise pull the latest changes
+# and check if there were any changes so we can skip rebuilding the venv if there weren't any
+if ! git clone https://github.com/canonical/checkbox.git; then
+    # Get the latest commit before pull
+    before_pull=$(git -C checkbox rev-parse HEAD)
+    echo "Before pull: $before_pull"
+
+    # Perform the pull
+    git -C checkbox pull
+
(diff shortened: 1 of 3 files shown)


Hook25 (approved):
Very nice that this includes a checkbox version in the log. Do you think a `pip freeze` could be useful as well? Not all requirements have a fixed version in the checkbox pyproject so maybe we could experience a failure due to an upgrade in a dependency?

kissiel:
> Very nice that this includes a checkbox version in the log. Do you think a `pip freeze` could be useful as well? Not all requirements have a fixed version in the checkbox pyproject so maybe we could experience a failure due to an upgrade in a dependency?

pip freeze where?

Hook25:
after installing checkbox (from within the venv) so that we get every package version in the log

plars (approved):
Looks fine, but wondering if it should really be part of checkbox itself rather than snap-update-tools? (in fact I thought it already had something similar to this)

kissiel:
> Looks fine, but wondering if it should really be part of checkbox itself rather than snap-update-tools? (in fact I thought it already had something similar to this)

It's hard to make it part of checkbox if this git clones checkbox :D

snap-update-tools/provision_checkbox.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+
+# This script is used to provision the controller node with the necessary dependencies
```
mz2, line 3: Please make a note about this script's existence in the top level README.

plars:
> > Looks fine, but wondering if it should really be part of checkbox itself rather than snap-update-tools? (in fact I thought it already had something similar to this)
> 
> It's hard to make it part of checkbox if this git clones checkbox :D

Sorry, I so the list of things that were getting apt installed and thought it was mostly about setting up everything needed for checkbox to run and thought it would be generally useful for people using checkbox. On second look I see that it's really just replicating the extra things we have in the job scripts that try to refresh the repo if it exists or create it if it's missing.

plars:
@kissiel based on the discussion we had yesterday, am I correct that the intention is to extend this so that you can specify a tag for the precise revision of the source to install?
What's the best way for us to find out which tag to use from what was installed on the the dut as a snap (for core systems) or deb (for DUTs where the snap is not used)

Hook25 (approved):
+1, ty

PR #4, opened 2023-10-08
add the kernel switching tool

kissiel:
# Ubuntu Kernel Switcher

## Summary

Ubuntu Kernel Switcher is a utility that lets you switch to a specific
kernel. Notably, post-switch, that kernel becomes the default, ensuring
it remains after system updates introduce new kernels.

## Usage

To run the tool, use invoke `switch_kernel.py` with the desired kernel
as the argument. Examples:

- `switch_kernel.py 6.5`
- `switch_kernel.py generic`
- `switch_kernel.py 6.2.0-33-generic`
- `switch_kernel.py realtime`

Note how the argument is fuzzy matched. But beware, first match will be used
so for instance if `realtime` is used as the argument, first grub's menuentry
that matches `realtime` will be used.

## Testing

There `test_in_lxd_vm.py` is a script that tests the tool inside LXD's VMs.
(So it's somewhat safe to be used).

Resolves: https://warthogs.atlassian.net/jira/software/c/projects/CER/boards/534?selectedIssue=CHECKBOX-904

### DIFF
--- a/kernel-switcher/switch_kernel.py
+++ b/kernel-switcher/switch_kernel.py
@@ -1,4 +1,17 @@
 #!/usr/bin/env python3
+# Copyright (C) 2023 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 """
 This program selects the kernel to boot into.
@@ -19,49 +32,31 @@ def get_submenu_entry(grub_cfg_contents: str):
     :param grub_cfg_contents: the contents of /boot/grub/grub.cfg
     :return: the identifier of the "advanced" submenu entry
     :raises SystemExit: if the identifier could not be found
-
-    >>> grub_entry = (
-    ... "submenu 'Advanced options for Ubuntu' $menuentry_id_option "
-    ... "'gnulinux-advanced-aca31037-7571-415c-b666-f565c524c2a6' {"
-    ... )
-    >>> get_submenu_entry(grub_entry)
-    'gnulinux-advanced-aca31037-7571-415c-b666-f565c524c2a6'
-    >>> get_submenu_entry('foo')
-    Traceback (most recent call last):
-        ...
-    SystemExit: Could not find submenu entry in grub.cfg
     """
 
     for line in grub_cfg_contents.splitlines():
-        if "submenu" in line:
+        if "submenu" in line and "Advanced" in line:
             identifier = line.split()[6]
             return identifier.replace("'", "")
     raise SystemExit("Could not find submenu entry in grub.cfg")
 
 
+def assert_root():
+    """
+    Do nothing if the user is root, otherwise exit the program wit han error.
+    """
+    if os.geteuid() != 0:
+        raise SystemExit("This program must be run as root.")
+
+
 def find_menuentry_for_kernel(keyword: str, grub_cfg_contents: str):
     """
     Returns the menuentry for the kernel that matches the keyword.
+
     :param keyword: the keyword to search for
     :param grub_cfg_contents: the contents of /boot/grub/grub.cfg
     :return: the menuentry for the kernel that matches the keyword
     :raises SystemExit: if the adequate menuentry could not be found
-    >>> grub_entry = (
-    ... "menuentry 'Ubuntu, with Linux 5.4.0-80-generic' --class ubuntu "
-    ... "--class gnu-linux --class gnu --class os $menuentry_id_option "
-    ... "'gnulinux-5.4.0-80-generic-advanced-aca31037-7571-415c-b666-"
-    ... "f565c524c2a6' {"
-    ... )
-    >>> correct_match = ('gnulinux-5.4.0-80-generic-advanced-'
-    ... 'aca31037-7571-415c-b666-f565c524c2a6')
-    >>> match = find_menuentry_for_kernel('5.4.0-80', grub_entry)
-    >>> match == correct_match
-    True
-    >>> find_menuentry_for_kernel('5.4.0-80', 'foo')
-    Traceback (most recent call last):
-        ...
-    SystemExit: Could not find kernel in grub.cfg
-
     """
 
     for line in grub_cfg_contents.splitlines():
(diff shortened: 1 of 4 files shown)


mz2 (commented):
Please include the usage and testing instructions in the (thus far completely empty, so I understand not very inviting) README.

mz2:
I filed a follow-up task for the next cycle (https://warthogs.atlassian.net/jira/software/c/projects/RTW/issues/RTW-166) to CI the LXD based test you added.

jocave (commented):
Thanks for the work, looks like could be very helpful. 

However, I find "switch to specific" and "fuzzy" to be somewhat oxymoronic. I'm imagining that I might want to use this in an SRU test for example, I want to switch to a kernel of my specification and fail noisily/hard if that kernel is not switched, rather than be helpfully switched to something that sounds nearly the same.

Does this help in that situation?

kissiel:
> Please include the usage and testing instructions in the (thus far completely empty, so I understand not very inviting) README.

Oh shoot, missing `git add`, I'll have access to the repo tomorrow. The readme content I wrote is almost the same as the description posted with this PR.

kissiel:
> Thanks for the work, looks like could be very helpful.
> 
> However, I find "switch to specific" and "fuzzy" to be somewhat oxymoronic. I'm imagining that I might want to use this in an SRU test for example, I want to switch to a kernel of my specification and fail noisily/hard if that kernel is not switched, rather than be helpfully switched to something that sounds nearly the same.
> 
> Does this help in that situation?

I agree, there's a tiny bit of contradiction there :D
FWIW, if you specify a concrete kernel, it's that kernel or deadsies, so it should fulfil your requirement. Fuzzy is a complete misnomer here - "concrete or partial" better describes it.
I guess that having said all of that you'd like to see something like `--also-match-partial` or using the inverse logic `--match-exact-name`. 
I wrote this tool with realtime kernel in mind, so of course I'm leaning towards the latter.

kissiel:
Added missing readme.

kernel-switcher/switch_kernel.py
```diff
@@ -0,0 +1,123 @@
...
+        if "menuentry" in line and keyword in line and "recovery" not in line:
+            # Searching for the pattern using a regular expression
+            match = re.search(r"\w*gnulinux.*\w", line)
+            if match:
+                return match.group()
+    raise SystemExit("Could not find kernel in grub.cfg")
+
+
+def main(argv):
+    parser = argparse.ArgumentParser(
```
Hook25, line 77: given how easy to implement and potentially useful it could be, can we have a dry-run option to get if a kernel parameter would work? (return after `"Found menuentry:..."`)

kernel-switcher/switch_kernel.py
```diff
@@ -0,0 +1,123 @@
...
+    print(f"Searching for menuentry for kernel {kernel}...")
+    menuentry = find_menuentry_for_kernel(kernel, grub_cfg_contents)
+    print(f"Found menuentry: {menuentry}")
+
+    # let's remove the "force partuuid" file from the grub config
+    # otherwise the kernel will not boot
+    print("Removing 'force partuuid' from grub config...")
+    file_path = "/etc/default/grub.d/40-force-partuuid.cfg"
+    if os.path.exists(file_path):
+        os.remove(file_path)
```
Hook25, line 104: from here onward we need to be `sudo`, can we check so that we avoid a traceback on this remove? If you add `--dry-run`, check here and not at the beginning of main

kernel-switcher/test_in_lxd_vm.py
```diff
@@ -0,0 +1,155 @@
...
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        print(f"Deleting '{self.vm_name}' VM...")
+        self._run_host_command(f"lxc delete --force {self.vm_name}")
+
+    def _run_host_command(self, cmd, supress=False):
+        """Executes a command on the host and returns the output."""
+        try:
+            cmd_list = shlex.split(cmd)
```
Hook25, line 58: Weird to use strings everywhere and split them here, why do we do this?

kernel-switcher/test_in_lxd_vm.py
```diff
@@ -0,0 +1,155 @@
...
+        if line.startswith("linux-image"):
+            extraceted_pkgs.append(line.split("/")[0])
+    return extraceted_pkgs
+
+
+def main():
+    interrupt_handler = InterruptHandler()
+    with VMContext("vm-test") as vm:
+        # Wait for VM to be ready
+        time.sleep(10)
```
Hook25, line 125: is this equivalent to `vm.wait_for_vm_ready(timeout=70)`

kernel-switcher/test_in_lxd_vm.py
```diff
@@ -0,0 +1,155 @@
...
+    extraceted_pkgs = []
+
+    for line in apt_log.splitlines():
+        if line.startswith("linux-image"):
+            extraceted_pkgs.append(line.split("/")[0])
+    return extraceted_pkgs
+
+
+def main():
+    interrupt_handler = InterruptHandler()
```
Hook25, line 122: duplicated?

kernel-switcher/test_in_lxd_vm.py
```diff
@@ -0,0 +1,155 @@
...
+        )
+        print("Running the kernel switcher...")
+        new_kernel = pkgs[0].replace("linux-image-", "")
+        vm.run_vm_command(f"python3 switch_kernel.py {new_kernel}")
+        print("Rebooting VM...")
+        vm.run_vm_command("reboot")
+        # we have to wait a few seconds to let the VM reboot
+        time.sleep(3)
+        print("Waiting for VM to come back...")
+        vm.wait_for_vm_ready()
```
Hook25, line 149: same as above, is this `vm.wait_for_vm_ready(63)`?

kernel-switcher/test_in_lxd_vm.py
```diff
@@ -0,0 +1,155 @@
...
+    # with just python's string methods
+    extraceted_pkgs = []
+
+    for line in apt_log.splitlines():
+        if line.startswith("linux-image"):
+            extraceted_pkgs.append(line.split("/")[0])
+    return extraceted_pkgs
+
+
+def main():
```
Hook25, line 121: I like this, but this is not a test in the pytest sense, is this intended?

kernel-switcher/test_in_lxd_vm.py
```diff
@@ -0,0 +1,155 @@
+#!usr/bin/env python3
```
Hook25, line 1: typo, also, this file is not executable (`no +x`), this is also probably an oversight

kernel-switcher/test_in_lxd_vm.py
```diff
@@ -0,0 +1,155 @@
...
+- reboot
+- check that the old kernel is _still not_ running
+- use the kernel switch tool to switch to the old kernel
+- reboot
+- check that the older kernel is running
+"""
+import subprocess
+import shlex
+import time
+import contextlib
```
Hook25, line 15: Minor: can be removed

kernel-switcher/test_in_lxd_vm.py
```diff
@@ -0,0 +1,155 @@
...
+
+class InterruptHandler:
+    def __init__(self):
+        self.interrupted = False
+        self.first_interrupt_time = time.time()
+        signal.signal(signal.SIGINT, self.signal_handler)
+
+    def signal_handler(self, signum, frame):
+        if self.interrupted and time.time() > self.first_interrupt_time + 2:
+            print("\nForced exit.You'll have to do the cleanup yourself.")
```
Hook25, line 27: Is this accurate? the cleanup is in the `__exit__` of the `VMContext`, shouldn't it cleanup? If not, could we use something like `atexit`?

kernel-switcher/test_in_lxd_vm.py
```diff
@@ -0,0 +1,155 @@
...
+            raise KeyboardInterrupt
+        self.interrupted = True
+        print(
+            "\nCTRL+C detected while using a VM...\n"
+            "Press CTRL+C again in 2 seconds if you really want to force exit"
+        )
+
+
+interrupt_handler = InterruptHandler()
```
Hook25, line 36: Why do we have this mechanism?

Hook25 (changes requested):
Overall nicely done, seems very useful.

Some suggestions here and there and a couple of questions

kernel-switcher/test_in_lxd_vm.py
kissiel: ah, thanks!

kernel-switcher/test_in_lxd_vm.py
kissiel: To handle cleanup gracefully, double tapping ctrl+c would kill the vm disposal.

kernel-switcher/test_in_lxd_vm.py
kissiel: but that cleanup could be interrupted by the user

kernel-switcher/test_in_lxd_vm.py
Hook25: I was thinking we could directly use lists everywhere

kernel-switcher/test_in_lxd_vm.py
kissiel: Yeah, it has nothing to do with unit testing. It's just a e2e test covering the whole flow.

kernel-switcher/test_in_lxd_vm.py
kissiel: Not fully equivalent... The time.sleep() is not an active wait so nothing is run and being printed. But with a testing tool like this it's not necessary to have super clean output.
Deleted.

yphus (approved):
+1

PR #5, opened 2023-10-19
get-snap-data: add support for monitoring named branches

awhitcroft:
The kernel team use stream2 branches to allow parallel development and testing.  Branches are intentionally not published in the main info data so use the refresh interface to interogate and report on specified branches.  For kernels assume stream2 by default.


awhitcroft:
I will note that this does not change the consumers of this data to cope with the proposed form.

PR #8, opened 2023-12-13
Get latest snap version

Hook25:
To create the last piece of the canary validation pipeline we need a way to communicate back the result of the validation to the Checkbox repository. We have decided to do so by moving the head of a branch called `beta_validation` to the latest validated commit in the repository. 

This PR introduces a small tool to gather what is the latest version of checkbox that was uploaded to the store. This information will be used to calculate the offset (using the `-devXX` suffix) and update the branch in a separate PR.

Minor: This uses a functionality that was implemented by another script, this functionality is therefore moved to an utils module shared between the two scripts. 

Related to CHECKBOX-1093 (together with https://github.com/canonical/hwcert-jenkins-tools/pull/9 resolves it)


snap-update-tools/get_version_store_snap.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def get_latest_version(snap_info, channel):
+    for entry in snap_info["channel-map"]:
+        if entry["channel"]["name"] == channel:
+            return entry["version"]
+    raise SystemExit("No version found in the specified channel")
+
+
+def get_version_store_snap(snap_name: str, channel: str):
```
kissiel, line 36: It takes a while to understand the function name.

The channel just flies through the function, and the caller already has whatever is returned, so I don't see the value in this arg.

kissiel (changes requested):
I'd really love a more precise name for the module that is now called the generic "utils".

snap-update-tools/get_version_store_snap.py
Hook25: The arg is there because that is the main functionality function. This is intended so that one is able to use the function without injecting+parsing the args. I have done this in all tools that I have developed for the release process, it is pretty convenient. You are right about the name, I have changed it, now it should flow way better!

kissiel (changes requested):
One small addition and one small fix proposed below.

kissiel (approved):
Yes, please! +1

PR #9, opened 2023-12-13
Script to move a target branch given a version string

Hook25:
To create the last piece of the canary validation pipeline we need a way to communicate back the result of the validation to the Checkbox repository. We have decided to do so by moving the head of a branch called beta_validation to the latest validated commit in the repository.

This introduces a script that is able to move the HEAD of a branch to the correct revision, calculating it from the latest tag and the provided version string. This is connected to: https://github.com/canonical/hwcert-jenkins-tools/pull/8.

The idea is that we first get the latest version in the store, then we use this script in a fresh clone of the Checkbox repository to move the tag like this:
```bash
$ git clone https://github.com/canonical/checkbox /path/to/repo
$ [...] # setup the credentials/repo identity
$ export CURRENT_VERSION="`./get_version_store_snap checkbox edge`"
$ ./move_branch_by_version.py /path/to/repo beta_validation $CURRENT_VERSION
```

Resolves CHECKBOX-1093

### DIFF
--- a/snap-update-tools/move_branch_by_version.py
+++ b/snap-update-tools/move_branch_by_version.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+"""
+This script moves a branch HEAD from the current location
+to an offset calculated from a version string in the form
+of vX.Y.Z-devAA, where AA is the ammount of commits
+
+Example usage:
+    python3 move_beta_branch.py repository_path beta_validation version_string
+"""
+
+import sys
+
+from argparse import ArgumentParser, ArgumentTypeError
+from subprocess import check_output, check_call
+
+
+def get_latest_tag(repo_path: str):
+    return check_output(
+        ["git", "describe", "--tags", "--abbrev=0"], cwd=repo_path, text=True
+    ).strip()
+
+
+def get_history_since(tag: str, repo_path: str):
+    return check_output(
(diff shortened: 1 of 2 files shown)


kissiel:
Either I'm missing something or this thing is completely off.

The version string that we plan to shift is never a tag. So the arg is off.
This program suggests it's a universal shifter, yet has some special stuff for beta.
Suggested use case has a huge race condition embedded inside of it (with the checking of the version from the store after it's been validated).
I struggle to understand why are we doing any shenanigans with the version strings, and computing anything.
The program here boils down to running three commands, and I can't see what value it brings, sorry.

I'm not selecting "request changes" because more appropriate would be to close this PR. I'm keeping it open for the potential discussion of things I may have missed.

kissiel:
Ah, now I get it. The version computation here is needed because we need to find the actual version that got marked as devXX.

It's starting to feel like we're introducing problems and potential problems onto ourselves here. I will strongly push towards having a ground truth, the commit hash, to be preserved and used for that.

ATM I think this could be reported by `checkbox-cli --version`, doesn't have to be a part of the version string, or any other means.

kissiel (approved):
We had a discussion with Max, where he explained to me why we need this whole thing instead of one smart git invocation.
 +1

PR #11, opened 2024-01-30
Merge wait for snaps and debs

fernando79513:
## Description

Merged the scripts to check if a specific snap and deb package is already published into "checkbox_version_published.py.

There will be a follow-up PR to remove the old separated scripts once the PR on "hwcert-jenkins-jobs" that includes these changes is merged 

## Resolved issues

[Merge wait for snaps and wait for debs scripts](https://warthogs.atlassian.net/browse/CHECKBOX-1216)

## Documentation

Added a readme to the "version_published" folder

## Tests
To run the tests:
```
cd version_published
pytest 
```


version-published/checkbox_version_published.py
```diff
@@ -0,0 +1,346 @@
...
+
+def get_package_specs(yaml_content: dict, version: str) -> list[PackageSpec]:
+    """
+    Create a list of PackageSpec objects from the yaml content.
+    :param yaml_content: file containing the package requirements
+    :param version: the version of the package
+    :return: the list of PackageSpec objects
+    """
+    version = version.replace("-", "~")
+    if "required-packages" not in yaml_content:
```
kissiel, line 212: we expect the yaml to be structured as a dict, so if we actually codify it (see below for details) then we can streamline the whole thing here to just do:

`yaml_content.get("required_packages", [])`
This way no need for branching.

Same goes for snaps ofc.

version-published/checkbox_version_published.py
```diff
@@ -0,0 +1,346 @@
...
+    )
+    parser.add_argument(
+        "--timeout",
+        help="Timeout in seconds after which the program will stop checking.",
+        default=300,
+        type=float,
+    )
+    args = parser.parse_args(argv[1:])
+
+    yaml_content = yaml.load(args.checkbox_yaml, Loader=yaml.FullLoader)
```
kissiel, line 338: If the yaml is empty, the loader will return None and the code will crash, it may be worth checking if it's a dict.

kissiel (changes requested):
TYVM! Almost there, some improvements suggestions inline

version-published/checkbox_version_published.py
```diff
@@ -0,0 +1,346 @@
...
+
+The specs are specified with a yaml that looks like this:
+    required-snaps:
+        - name: checkbox22
+            channels: [ latest/edge ]
+            architectures: [ amd64, arm64 ]
+
+So the invocation of this program would be:
+    python3 checkbox_version_published.py 3.3.0-dev10 \
+    --snaps checkbox-canary-snaps.yaml
```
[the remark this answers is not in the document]
fernando79513: Yes, I forgot to change the docstring, I will also update it

fernando79513:
I applied the comments you suggested. I also modified a bit the structure of the packages yaml so its more similar to the snaps one, moving the `channel` to each package, so I don't have to deal with a different lists/dicts structure.
I also updated the tests accordingly.

kissiel (approved):
Great stuff, thanks for all the tweaks!

PR #12, opened 2024-02-19
Removed old wait for snaps and debs

fernando79513:
Since we have merged both scripts, we can remove now the `ppa-update-tools` and the `snap-update-tools` folders


kissiel (approved):
It looks that all the replacing bits are in place.
Fantastic conclusion!
+1

PR #15, opened 2024-02-21
Enable efi-vars from the kernel switcher

kissiel:
This patch introduces an arg to the kernel switcher that, when used, places `efi=runtime` kernel command line argument in the grub config. This value is necessary for the device to be provisionable by MAAS, and when the realtime kernel is used, that option is disabled (the default value for non-realtime kernels is indeed `efi-runtime`).

### DIFF
--- a/kernel-switcher/switch_kernel.py
+++ b/kernel-switcher/switch_kernel.py
@@ -148,7 +148,6 @@ def parse_args(argv):
 
     return parser.parse_args(argv[1:])
 
-    return args.kernel[0], args.dry_run, args.enable_efi_vars
 
 def add_efi_opt(cmdline):
     """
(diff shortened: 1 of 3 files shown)


kernel-switcher/switch_kernel.py
```diff
@@ -140,12 +140,55 @@ def parse_args(argv):
...
 
-    args = parser.parse_args(argv[1:])
-    return args.kernel[0], args.dry_run
+    parser.add_argument(
+        "--enable-efi-vars",
+        action="store_true",
+        help="enable EFI variables so MAAS can reprovision the machine",
+    )
+
+    return parser.parse_args(argv[1:])
```
plars, line 149: I don't think the argv[1:] is actually required, I've used this before to make my parse_args() function easier to test, and it always seems fine to just pass argv to it. Also useful for avoiding typeerror on the off chance that argv==None (which is technically the default but unlikely, yes).

plars (approved):
Looks ok to me, but a couple of suggestions that you can feel free to take or leave.

One minor one below about the args. But the other one is about the large number of patches on your tests for main. Trying to get coverage in your tests for main is problematic like that, but I wonder if you can make this nicer to read/use by injecting a dependency here to handle the details? Then you can just create a "test" version of what you inject that let's you confirm the right things have been called, without having to patch all the things.

kernel-switcher/switch_kernel.py
```diff
@@ -140,12 +140,55 @@ def parse_args(argv):
         help="do not write anything to the grub config",
     )
 
-    args = parser.parse_args(argv[1:])
-    return args.kernel[0], args.dry_run
+    parser.add_argument(
```
Hook25, line 143: Question: why this an `enable-XX`? If this is not specified the machine will be un-reprovisionable via MAAS if switching to a real time kernel, is there a situation where we don't want this?

Hook25 (commented):
I have a couple of questions. I'm a bit worried that I can't see this kernel option in the docs, I may be missing something here

https://www.kernel.org/doc/html/v6.1/admin-guide/kernel-parameters.html

kernel-switcher/switch_kernel.py
kissiel: Yes, we don't need it on projects handled by other methods of provisioning, where indeed efi runtime might affect realtime-ness.

kissiel:
I'm landing this to have a closure on a needed feature while I refactor the main function so it's not insane.

Hook25 (approved):

PR #16, opened 2024-03-08
Update provision_checkbox.sh to fully support any REF (checkbox version)

Hook25:
The `provision_checkbox.sh` script assumes that the version that we need is tagged. We have changed this design so now it wont be the case for any beta/candidate we provision. This changes the script to support this new behaviour where we can provision any valid Checkbox version.

Examples:
```
(venv)  tmp >  ~/prj/canonical/hwcert-jenkins-tools/provision_checkbox.sh v3.3.0-dev1
[...]
++ checkbox-cli --version
$PROVIDERPATH is defined, so following provider sources are ignored ['/usr/local/share/plainbox-providers-1', '/usr/share/plainbox-providers-1', '/home/h25/.local/share/plainbox-providers-1', '/var/tmp/checkbox-providers-develop'] 
+ echo checkbox version 3.3.1.dev1+g4114cb70c
checkbox version 3.3.1.dev1+g4114cb70c
(venv)  tmp >  ~/prj/canonical/hwcert-jenkins-tools/provision_checkbox.sh v3.3.0
[...]
++ checkbox-cli --version
$PROVIDERPATH is defined, so following provider sources are ignored ['/usr/local/share/plainbox-providers-1', '/usr/share/plainbox-providers-1', '/home/h25/.local/share/plainbox-providers-1', '/var/tmp/checkbox-providers-develop'] 
+ echo checkbox version 3.3.0
checkbox version 3.3.0
```

### DIFF
--- a/version-published/snap_info_utility.py
+++ b/version-published/snap_info_utility.py
@@ -39,6 +39,8 @@ def get_history_since(tag: str, repo_path: str):
 
 
 def get_offset_from_version(version: str) -> int:
+    if "dev" not in version:
+        return 0
     return int(version.rsplit("dev", 1)[1])
 
 
--- a/version-published/snap_info_utility.py
+++ b/version-published/snap_info_utility.py
@@ -31,7 +31,7 @@ def get_history_since(tag: str, repo_path: str):
             "log",
             "--pretty=format:%H",
             "--no-patch",
-            f"{tag}..origin/main",
+            f"{tag}~1..origin/main",
         ],
         text=True,
         cwd=repo_path,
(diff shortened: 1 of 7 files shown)


version-published/snap_info_utility.py
```diff
@@ -20,3 +22,39 @@ def get_snap_info_from_store(snap_name: str) -> dict:
...
+
+
+def get_revision_at_offset(version: str, repo_path: str):
+    tag = get_latest_tag(repo_path)
+    history = get_history_since(tag, repo_path)
+    offset = get_offset_from_version(version)
+    # history is now -> tag(inclusive)
+    # raw tag (no dev) is tag so -0 -1 = -1 (latest in history)
+    # dev-1 is tag+1 so -1 -1 = -2 semi lastest in history
+    return history[-offset - 1]
```
kissiel, line 54: This smells unsafe.

history IIUC is a list of strings.
I understand that offset is positive, this is why you use a unary `-`. But in the end there's a lot of opportunities here to get a negative index. 

I'm guessing the correctness can be inferred from the relationships between functions, but this is true only "right now". Boundary checks here would make it easy on the next person. Or assertions.

kissiel (changes requested):
Requesting more safety below.

version-published/snap_info_utility.py
kissiel: Also the comment here suggests that the index should be negative ? :confused:

Hook25: offset is positive (as it is the distance from the tag). Given that we have an history from now to the tag, we have to use a negative index to "add" to the tag as:
- history[-1] is the tag
- history[-2] is the commit after the tag
- hitory[0] is the current HEAD

I will add the boundary checking now

PR #18, opened 2024-03-19
Delete data directory

jocave:
Removing files that were used solely to support trello dashboard.

Deleting this data as [confirmed by previous owner](https://chat.canonical.com/canonical/pl/n9jgt98giinozenp8phgf4oixh) that is was only for the trello workflow which has now been replaced


plars (approved):
+1

PR #21, opened 2024-04-16
Implement scriptlet to fetch manifests from remote server

andrejvelichkovski:
## Description
This scriptlet implements a simple network request that will either fetch a manifest file or provide a default one for SRU testing.

## Testing
This change is manually tested by calling `fetch_manifest <CID> <SECURE_ID>`


scriptlets/sru-helpers/fetch_manifest
```diff
@@ -0,0 +1,69 @@
...
+#
+# Usage:
+#     `fetch_manifest <CID> <MACHINE_SECURE_ID>`
+
+from argparse import ArgumentParser
+from io import StringIO
+import requests
+
+
+# This is the manifest that all machines will use if we cannot find
```
boukeas, line 15: If the purpose of this script is to retrieve a manifest using the API, then I don't think the script itself is the best place to define the default manifest. Perhaps it should be in a separate file but my suggestion would be for the API itself to return the default manifest.

scriptlets/sru-helpers/fetch_manifest
```diff
@@ -0,0 +1,69 @@
...
+    manifest_conf = StringIO()
+    print("        [manifest]", end="\n", file=manifest_conf)
+    for key, value in manifest_json.items():
+        line = f"        {key} = {value}"
+        print(line, end="\n", file=manifest_conf)
+    return manifest_conf.getvalue()
+
+
+def fetch_manifest():
+    parser = ArgumentParser(
```
boukeas, line 45: You might as well fill these arguments in!

scriptlets/sru-helpers/fetch_manifest
```diff
@@ -0,0 +1,69 @@
...
+        prog="",
+        description="",
+    )
+    parser.add_argument("cid", type=str)
```
boukeas, line 49: Good to also add `help` arguments here.

scriptlets/sru-helpers/fetch_manifest
```diff
@@ -0,0 +1,69 @@
...
+        com.canonical.certification::has_thunderbolt3 = false
+        com.canonical.certification::has_touchscreen = false
+        com.canonical.certification::has_tpm2_chip = false
+        com.canonical.certification::has_usb_storage = true
+        com.canonical.certification::has_usb_type_c = false
+        com.canonical.certification::has_wlan_adapter = true
+"""
+
+
+def parse_json(manifest_json):
```
boukeas, line 32: I am pretty sure you can use `ConfigParser` for this sort of thing. https://docs.python.org/3/library/configparser.html

scriptlets/sru-helpers/fetch_manifest
```diff
@@ -0,0 +1,69 @@
...
+        f"https://certification.canonical.com/api/v2/machines/{args.cid}/manifest?secure_id={args.secure_id}"
+    )
+    if response.status_code == 200:
+        print("Found manifest")
+        manifest_conf = parse_json(response.json())
+    else:
+        print("Manifest not found, using default")
+        manifest_conf = fallback_manifest
+
+    with open("manifest_conf", "w") as f:
```
boukeas, line 64: Perhaps it would be convenient to make the name of the output file a command-line argument as well (with a reasonable default).

boukeas (changes requested):

scriptlets/sru-helpers/fetch_manifest
andrejvelichkovski: Good idea! I have added this optional argument now

scriptlets/sru-helpers/fetch_manifest
andrejvelichkovski: Oh, while rushing to finish up this script, I have forgotten to focus on the code readability. Thanks for pointing this out, I have added these.

scriptlets/sru-helpers/fetch_manifest
andrejvelichkovski: I have added help definitions for all arguments now.

scriptlets/sru-helpers/fetch_manifest
andrejvelichkovski: Seems like `ConfigParser` is more powerful than I expected it to be. I have tried to incorporate all built-in methods in the script.

scriptlets/sru-helpers/fetch_manifest
```diff
@@ -0,0 +1,75 @@
...
+fallback_manifest_config = configparser.ConfigParser(delimiters=("=",))
+fallback_manifest_config.read_string(fallback_manifest_raw)
+
+
+def parse_json(manifest_json):
+    """
+    Given an manifest json file, parse it as a format suitable for the Checkbox launcher file
+    """
+    manifest_config = configparser.ConfigParser(delimiters=("=",))
+    manifest_config["manifest"] = manifest_json
```
boukeas, line 40: I think it would make sense if the endpoint reply wasn't a shallow dict with all the key-value pairs but rather a dict that contained the `manifest` field you are now manually adding, along with the key-value pairs underneath that. This is would also allow you to add extra fields (other than the `manifest`) to the endpoint's response at a later point, if required.

scriptlets/sru-helpers/fetch_manifest
boukeas: It would still be useful if the endpoint (assuming it's reachable) returned a default manifest if the requested one wasn't located. Or perhaps there could be a variation of the endpoint that you would hit just to retrieve a default manifest, in case the original request was unsuccessful.

boukeas: I would also argue that this script shouldn't return a hardwired default manifest in any case. It's purpose is to retrieve a manifest from the endpoint. If the endpoint is unreachable then the script exits with an error code because it has been unsuccessful in fulfilling its purpose.

boukeas (changes requested):

scriptlets/sru-helpers/fetch_manifest
andrejvelichkovski: I have updated the script to exit with exit code 1 in case it cannot fetch the manifest. The user of the script can then from the exit code whether the manifest is fetched or not and use the default instead.

scriptlets/sru-helpers/fetch_manifest
andrejvelichkovski: Following the discussion on MM, I think we should keep the simple structure for now, until we have a large collection of manifests before we actually start adding more features on top of the manifest data.

boukeas (approved):

PR #22, opened 2024-04-24
add wait-for-apt script

kissiel:
## Scriptlet that waits for apt and dpkg

The original inline shell code had multiple problems:
- on timeout it killed every process that had `apt` **anywhere in the path of the process originating binary** 
- **always** reported that apt was running even though it was not
- triple-backslashes and the enforced indentation due to being interpolated into a job definition
- IMHO wasn't easy to read and understand what it did

Previous implementation:
https://github.com/canonical/hwcert-jenkins-jobs/blob/main/jobs/upgrade-testing/wait-resource.sh

### DIFF
--- a/scriptlets/wait_for_apt_dpkg_resource
+++ b/scriptlets/wait_for_apt_dpkg_resource
@@ -3,25 +3,43 @@
 # If times out, kills those processes.
 
 is_process_running() {
-    # the arg is the process name
-    pgrep "$1" > /dev/null
+    # The arg is the process name
+    pgrep "$1" > /dev/null 2>&1
 }
-wait_for_apt_dpkg_resource_function() {
 
+is_file_used() {
+    # The arg is the file name
+    fuser -s "$1"
+}
+
+is_resource_busy() {
+    if \
+        is_process_running apt || \
+        is_process_running dpkg || \
+        is_file_used /var/lib/apt/lists/lock || \
+        is_file_used /var/lib/dpkg/lock || \
+        is_file_used /var/cache/debconf/config.dat;
+    then
+        return 0
+    else
+        return 1
+    fi
+}
+
+wait_for_apt_dpkg_resource_function() {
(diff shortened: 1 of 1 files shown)


plars (commented):
Thank you for doing this, it would be great to get this consolidated into **one** way of doing it and putting it in a scriptlet like this!

Not sure if you were ready for review on it yet but I had a few comments and considerations if you're interested:

Finding "when" it's ok to run apt and/or preventing it from automatically running is notoriously difficult.  IIRC the original script this came from also had to do some thing like finding processes that were holding /var/lib/apt/lists/lock, /var/lib/dpkg/lock, and /var/cache/debconf/config.dat with fuser then also making sure those processes had completed.

It might be worth looking into if that's needed too.

boukeas:
I have just started [this task](https://warthogs.atlassian.net/browse/CERTTF-329) and happened to stumble onto this PR in this context. @kissiel, the implementation used for the SRU jobs is [here](https://github.com/canonical/hwcert-jenkins-jobs/blob/main/jobs/sru/wait-resource.sh) and is slightly more refined: like @plars mentions it adds `fuser` to the mix and checks that certain lock files are not linked to any processes.

scriptlets/wait_for_apt_dpkg_resource
```diff
@@ -0,0 +1,33 @@
...
+wait_for_apt_dpkg_resource_function() {
+
+    # Get the current datetime in seconds since epoch
+    current_time=$(date +%s)
+
+    # Calculate the future datetime (10 minutes from now) in seconds since epoch
+    deadline=$(( $current_time + 600 ))
+
+    # Loop until there are no apt or dpkg processes or timeout occurs
+    while is_process_running "apt" || is_process_running "apt-get" || is_process_running "dpkg"; do
```
boukeas, line 18: I think that `apt-get` might be redundant here. `pgrep apt` should also cover that case.

PR #23, opened 2024-05-03
Adds test executions rerunner script

omar-selo:
A Jenkins job will fetch this script then run it to rerun jobs requested by TO. This is required for [RTW-303](https://warthogs.atlassian.net/browse/RTW-303)

[RTW-303]: https://warthogs.atlassian.net/browse/RTW-303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scriptlets/test-executions-rerunner/test_executions_rerunner.py
```diff
@@ -0,0 +1,71 @@
...
+            rerun_request["ci_link"]
+        )
+        if base_job_link:
+            family = rerun_request["family"]
+            if family == "deb":
+                self._submit_deb_rerun(base_job_link)
+            elif family == "snap":
+                self._submit_snap_rerun(base_job_link)
+            else:
+                logging.error(f"Invalid family name {family}")
```
nadzyah, line 50: Alternatively, we can use match/case instead here (or a dict)

nadzyah (approved):

scriptlets/test-executions-rerunner/tox.ini
```diff
@@ -0,0 +1,21 @@
+[tox]
+minversion = 4
+envlist = py38
```
plars, line 3: Seems to rely on having py38 on your system? Is there a way to have tox ensure this exists? If not, then maybe it's better to have it use the system installed python and use a github workflow to ensure we have the right version? Otherwise just running tox here doesn't work

PR #24, opened 2024-05-03
Fix test observer api url and adds GH testing workflow

omar-selo:
This is a followup to last PR https://github.com/canonical/hwcert-jenkins-tools/pull/23

Turns out there was a bug with test observer api url. So this PR fixes that. I can also confirm that it's [working](http://10.102.156.15:8080/job/test-executions-rerunner/23/console).

Additionally, I've added a GH workflow to test all scriptlets as suggested https://github.com/canonical/hwcert-jenkins-tools/pull/23#discussion_r1589108408. You can confirm that it's working from here https://github.com/canonical/hwcert-jenkins-tools/actions/runs/8939122569

### DIFF
--- a/.github/workflows/test_scriptlets.yaml
+++ b/.github/workflows/test_scriptlets.yaml
@@ -1,0 +1,34 @@
+name: Test Scriptlets
+on:
+  push:
+    branches-ignore:
+      - 'main'
+    paths:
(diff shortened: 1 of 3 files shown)


plars (approved):
Nice, more testing too!

PR #27, opened 2024-05-26
Remote install checkbox snaps

plars:
This resolves part of LM-1571
Basically, I wanted to make a reusable way of ensuring we get the right combination of checkbox and checkbox runtime snap installed. Sometimes we can figure that out based on the track/risk of checkbox, but other times, it will require knowing the checkbox runtime snap that we want such as with oem checkbox versions.

I don't have the associated jenkins template changes for this just yet, but I did a trial run of using it instead of the traditional install methods at http://10.102.156.15:8080/job/cert-rpi5b8g-arm64-uc24-snapd-beta/13/ in order to get the beta version of checkbox, and it installed the expected versions. Here are the log messages from it:
```
15:59:18 + hwcert-jenkins-tools/remote_install_checkbox_snaps/remote_install_checkbox_snaps.py --user ubuntu --remote 10.102.163.131 --checkbox-snap checkbox --checkbox-channel beta --checkbox-track uc24 --checkbox-args=--devmode --checkbox-runtime=
15:59:18 INFO:root:Running 'sudo snap install checkbox24 --channel=latest/beta' on 10.102.163.131
15:59:18 INFO:root:Executing command on 10.102.163.131: sudo snap install checkbox24 --channel=latest/beta
16:00:43 INFO:root:Running 'sudo snap install checkbox --channel=uc24/beta --devmode' on 10.102.163.131
16:00:43 INFO:root:Executing command on 10.102.163.131: sudo snap install checkbox --channel=uc24/beta --devmode
```


remote_install_checkbox_snaps/test_remote_install_checkbox_snaps.py
omar-selo, line 1: Can you add an entry to run these tests automatically in [test_scriptlets.yaml](https://github.com/canonical/hwcert-jenkins-tools/blob/main/.github/workflows/test_scriptlets.yaml)

omar-selo (commented):
Looks good to me, just a comment

remote_install_checkbox_snaps/remote_install_checkbox_snaps.py
```diff
@@ -0,0 +1,150 @@
...
+
+class SSHClient:
+    def __init__(self, host: str, username: str):
+        self.host: str = host
+        self.username: str = username
+
+    def execute_command(self, command: str):
+        logging.info("Executing command on %s: %s", self.host, command)
+        try:
+            ssh_command = (
```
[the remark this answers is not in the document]
plars: True, and I agree these are probably options that we want to add here especially since we use them everywhere by default. However I don't think we can depend on the _run macro for this.

remote_install_checkbox_snaps/test_remote_install_checkbox_snaps.py
plars: Does that imply that it should go under the scriptlets dir then? I thought that dir was only for the bite-sized bash macros that get reused by the jobs?

remote_install_checkbox_snaps/remote_install_checkbox_snaps.py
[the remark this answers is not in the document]
plars: I do see what you mean, but I think it might limit the utility of having this as a reusable script. Having it also handle the installation means that we can replace a good chunk of the scripting and templating for this in the jenkins job with just a command that we pass a few arguments to. In reality, we run everything with those same ssh args, so I don't see much need for them to change.

I'm curious if @jocave also has any thoughts on this, since I think it was his suggestion to put this in a reusable script?

remote_install_checkbox_snaps/remote_install_checkbox_snaps.py
```diff
@@ -0,0 +1,150 @@
...
+        track_digits: str = "".join(
+            filter(str.isdigit, self.options.checkbox_track)
+        )[:2]
+        return f"checkbox{track_digits}"
+
+    def get_checkbox_runtime_install_cmd(self) -> str:
+        """Get the install command for the checkbox runtime snap"""
+        return (
+            f"sudo snap install {self.get_checkbox_runtime()} "
+            f"--channel=latest/{self.options.checkbox_channel}"
```
boukeas, line 81: In [`run-sru.sh`](https://github.com/canonical/hwcert-jenkins-jobs/blob/6d3bc41c557a8c8b110857f1c85ecc99029d49db/jobs/sru/run-sru.sh#L224C13-L224C66) the channel is always `beta`. Here, it is assumed that the channel will be provided (and, from the examples, it doesn't look like it's always going to be `beta`). What's the logic that determines what the channel should be? Does it need to be captured in this tool or externally?

plars: I think for deb SRUs, that's probably always beta. However it's not universally true that we always want beta. For instance, the checkbox team may want to install edge or candidate in some of their tests. The main example is snaps, where we probably want beta (hence the whole reason for needing something like this), but for oem checkbox snaps, we don't always have the option of installing it from beta as the OEMs control the brand stores. Many of those checkbox snaps need to come from stable or edge, and this is already captured in the `{checkbox-channel}` value for those jobs.

remote_install_checkbox_snaps/test_remote_install_checkbox_snaps.py
omar-selo: Not necessarily. I just figured if we have tests to run, why not run them in CI

boukeas:
I think this can be closed, the `install_checkbox_snaps` script in https://github.com/canonical/hwcert-jenkins-tools/pull/53 takes care of determining the runtime version automatically.

PR #28, opened 2024-06-03
feat: Add downloadable installer for this repo

boukeas:
Certification jobs that need to make use of the tools in this repo will currently explicitly clone it, often in inconsistent ways:
```
git clone --depth=1 https://github.com/canonical/hwcert-jenkins-tools.git > /dev/null
git -C hwcert-jenkins-tools pull -q || (rm -rf hwcert-jenkins-tools && git clone https://github.com/canonical/hwcert-jenkins-tools.git)
```

This PR introduces an installer that can handle cloning this repo in a readable consistent manner. It can also clone from branches, useful in cases where new tools are being developed and tested:
```
curl -Ls -o installer.sh https://raw.githubusercontent.com/canonical/hwcert-jenkins-tools/main/installer.sh
source installer.sh tools --branch LM-1580-sru-refactor
```
Sourcing the script instead of executing it also defines the `add_to_path` function:
```
add_to_path tools/scriptlets
```

Here is an example of using the downloadable installer from within [a Testflinger job](https://testflinger.canonical.com/jobs/5f2b5b0f-be17-4c96-b9b6-8d7d817c0422).

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+
+# Clone the certification tools repo to a local directory.
+# If the repo is already available locally, fetch the latest version.
+# Use the --branch option to specify a specific branch
+
+
+fetch() {
+    if [ "$#" -lt 1 ]; then
+        echo "Usage: ${FUNCNAME[0]} <path> [<branch>]"
+        echo "Error: You need to provide a path to a local git repo."
+        return 1
+    fi
+    local LOCAL=$1
+    local BRANCH=${2:-main}
+    git -C "$LOCAL" fetch -q --update-head-ok origin $BRANCH:$BRANCH 2> /dev/null && \
+    git -C "$LOCAL" checkout -q $BRANCH && \
+    echo "Fetched branch $BRANCH into local repo $LOCAL"
+}
+
+
+clone() {
+    if [ "$#" -lt 2 ]; then
+        echo "Usage: ${FUNCNAME[0]} <repo> <path> [<branch>]"
+        echo "Error: You need to provide a git repo and a path to clone into."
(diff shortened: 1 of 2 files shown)


Hook25 (changes requested):
This looks super useful and well done. Please copy paste the description to this PR into a README.md file in the repo root

Hook25 (approved):

installer.sh
```diff
@@ -0,0 +1,82 @@
...
+
+# Clone the certification tools repo to a local directory.
+# If the repo is already available locally, fetch the latest version.
+# Use the --branch option to specify a specific branch
+
+
+fetch() {
+    if [ "$#" -lt 1 ]; then
+        echo "Usage: ${FUNCNAME[0]} <path> [<branch>]"
+        echo "Error: You need to provide a path to a local git repo."
```
plars, line 11: Does path really need to be provided or can we just have a safe default for it to use the repo name or "tools" or something? I like that you can specify a branch, but I suspect just sourcing this with no params would be sufficient in most cases.

installer.sh
```diff
@@ -0,0 +1,82 @@
...
+            else
+                echo "Error: Invalid argument $1"
+                usage
+            fi
+            ;;
+    esac
+    shift
+done
+
+TOOLS_REPO=https://github.com/canonical/hwcert-jenkins-tools.git
```
plars, line 79: Can we put all these global defaults at the top so they are easy to find when looking at the file?

plars (commented):
Really nice, just a few small suggestions below and one more here:
can we change the name from installer.sh to something that makes it clear what we're installing? Maybe `install-cert-tools.sh` or something?

PR #30, opened 2024-06-07
fix(scriptlets): modify `SSH_OPTS` to allow `_put` to work

boukeas:
The current version of `set-ssh-options` includes a `-l` option in the `SSH_OPTS`. This specifies the user when used with `ssh` but it specifies a limit when used with `scp`. As a result, the `_put` scriptlet (which relies on `scp`) is broken. 

This PR removes the `-l` option from `SSH_OPTS`. It also modifies the scriptlets that use `SSH_OPTS` accordingly, i.e. by specifying the user alongside the device IP.

Note that `set-ssh-options` is renamed to `set_ssh_options` for consistency. (I personally much prefer dashes but the idea is that scriptlets will be a drop-in replacement for inline functions -- `set-ssh-options` is used nowhere explicitly, whereas the scriptlets with underscores in their names are).

This [Testflinger job](https://testflinger.canonical.com/jobs/f20968bc-2158-4088-ae63-d30f5c8bb413) illustrates how the current version of `_put` does not work while the modified one from this branch does. It also illustrates that `_run` and the other scriptlets modified by this PR are still functional.


boukeas:
I am at a loss to make the `wait_for_snap_complete` test to work. It mocks `ssh` but `wait_for_snap_complete` uses `_run` and the mock `ssh` never runs. To be perfectly honest, I am not a huge fan of these BATS tests.

Hook25 (approved):
ty for the fix, +1

PR #36, opened 2024-07-03
[CERTTF-337] feat: Re-implement and extend scriptlets

boukeas:
# Description

From [CERTTF-337](https://warthogs.atlassian.net/browse/CERTTF-337):

> At the moment, certification scripts in the [hwcert-jenkins-jobs repo](https://github.com/canonical/hwcert-jenkins-jobs) define essential inline functions such as `_run`, `_put`, `wait_for_ssh`, `wait_for_resource`, that are then used throughout each job.
>
> At the same time, the public [hwcert-jenkins-tools repo ](https://github.com/canonical/hwcert-jenkins-tools) offers a range of software tools that include (but are not limited to) a set of “scriptlets” that were meant to replace and extend the essential functions described above.
>
> The current situation is that certification scripts in the jobs repo clone the tools repo and use _some_ of the complex tools available there (e.g. syncing the Checkbox version between the agent and the DUT) but they do not currently use any of the essential scriptlets.

This PR re-implements the existing essential scriptlets and introduces some additional ones to be used within SRU jobs (see the Changelog for details). Please refer to the [companion PR](https://github.com/canonical/hwcert-jenkins-jobs/pull/495) in the `hwcert-jenkins-job` repo to see how these scriptlets are used in practice in the context of SRU jobs.

## Changelog

- Existing scriptlets comprise a single main function and boilerplate code that only runs the function if the code is executed (rather than sourced). The idea probably was that other scriptlets would selectively source the functions they require but this is really not scalable when the dependency tree gets deeper: sourcing a function would also require sourcing all its dependencies. This PR does away with this structure for scriptlets: almost all scriptlets are meant to be executed, not sourced. Also, any directory containing scriptlets should be in the `PATH` (so that it is possible for higher-level scriptlets to execute the ones they depend on). This is taken care of by the installer (i.e. `install_tools.sh`).
- Barrier scriptlets: Certification scripts often use functions such as `wait_for_ssh` (wait until we can ssh into a device), `wait_for_snap_complete` (wait until all snap operations are complete) or `wait_for_resource` (wait until all package operations are complete). The current implementations are hard to follow because they mix the conditions, i.e. what they check in order to determine whether or not they should keep waiting, with the retry logic. This PR separates the two: there are checking scripts that _only_ describe the conditions (e.g. `check_for_ssh`) and waiting scripts that repeatedly call the checking scripts. For this purpose, a `retry` scriptlet has been introduced in order to capture the repetition logic and simplify the code. It follows the logic of the [`retry` tool ](https://manpages.ubuntu.com/manpages/mantic/man1/retry.1.html) while avoiding the dependency.
- The installer not only fetches the scriptlets but also adds them to the path and installs their dependencies so that they "just work". The installer also copies over some of the scriptlets that are required on the device, taking care to move them to a directory that is already in the `PATH` (which is more straightforward than modifying the `PATH` on the device).
- The new `stacker` tool is introduced, in order to handle the assembly of the Checkbox configuration file.
- The new scriptlet `install_checkbox_agent_source` handles the installation of Checkbox on the agent from source, synced to a version that matches the one installed on the device.
- The new scriptlet `install_packages` handles the installation of debian packages

### Testing with BATS

This PR removes testing of the scriptlets with BATS. Testing even the simplest scriptlets involved redefining e.g. `ssh` with functions in order to mock its behaviour. I don't feel that constitutes real testing and I don't see how it could scale to handle different situations or larger, more higher-level scriptlets that depend on calling other scriptlets.

Having said that, I can't offer an alternative at the moment other than local manual unit testing and integration testing by running actual certification jobs. In my view, the lack of unit testing can only be remedied if the scriptlets are implemented as a Python package.

## Testing

See the [companion PR](https://github.com/canonical/hwcert-jenkins-jobs/pull/495) which includes SRU test runs that make use of all these new features.

[CERTTF-337]: https://warthogs.atlassian.net/browse/CERTTF-337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scriptlets/install_checkbox_agent_source
```diff
@@ -0,0 +1,49 @@
...
+fi
+
+# clone Checkbox repo
+git clone --filter=tree:0 https://github.com/canonical/checkbox.git > /dev/null
+
+# checkout commit to match the required version
+$TOOLS_PATH/version-published/checkout_to_version.py ~/checkbox "$VERSION"
+
+# install checkbox from source
+(cd checkbox/checkbox-ng; sudo python3 setup.py install > /dev/null)
```
plars, line 48: I think the original version of this ensured that checkbox gets installed in a virtualenv. Do we ensure that somehow?

install_tools.sh
```diff
@@ -55,25 +55,31 @@ clone() {
...
+
+# fuser is required by `check_for_packages_complete`
+# (so install it on both the agent and the device)
+$SCRIPTLETS_PATH/defs/install_psmisc
+cat $SCRIPTLETS_PATH/defs/install_psmisc | _run bash
+[ $? -eq 0 ] || exit 1
+
+# install launcher tool on the agent
 add_to_path ~/.local/bin
+pip -q install $TOOLS_PATH/cert-tools/launcher
```
plars, line 85: Does this get installed in a virtualenv? it might accidentally work for now because the agent host container we use is based on focal I think, but if this ever gets updated to more recent versions, then it will fail.

scriptlets/_put
```diff
@@ -1,22 +1,43 @@
...
-#   `_put <local_file> <remote_file>`
-#   Or source it in another script and call
-#   `put_function <local_file> <remote_file>`
+# The IP of the device is provided by the environment variable DEVICE_IP,
+# which must be set. The user name on the remote device is `ubuntu`
+# by default, unless provided by the environment variable DEVICE_USER.
+#
+# The last argument provided is considered to be the "target", i.e. where
+# on the remote device the file(s) will be copied. If the last argument is
+# `--`, then the file(s) will be copied on the home directory of DEVICE_USER
```
plars, line 13: why `--` to indicate /home/ubuntu? Does this help in cases where we don't necessarily know what the default user would be? I know that some devices use "admin" or "testuser" for example, but I think we will need the default user somehow for those though, don't we?
I only ask because this is a little opaque, and not clear where it's getting copied to when I saw it in use.

plars (commented):
I had a quick look and had a few questions. Looks great so far and thanks for linking to an sru job where you made use of this, that helps a lot!

Also I'm curious what the reason was for removing the bats tests. You said that it's probably less than adequate (I think you're right, but that's just one way that shell scripts can be very limiting). But since we don't really have a better option right now, is it better to leave it in until we do?

scriptlets/install_checkbox_agent_source
boukeas: This is [corresponding code](https://github.com/canonical/hwcert-jenkins-jobs/blob/436cf72337a4bcfec2b42215361769d00477118f/jobs/sru/run-sru.sh#L325) in `run-sru.sh`:
```
hwcert-jenkins-tools/version-published/checkout_to_version.py ~/checkbox "\$CHECKBOX_VERSION"
(cd checkbox/checkbox-ng; sudo python3 setup.py install > /dev/null)
```
It's practically identical and I don't see a virtualenv being used. We can, of course, and that's why I wanted to offer an abstraction over the original code.

install_tools.sh
boukeas: No, I can change it so that it does. But can you just explain why it will fail on more recent versions?

plars: There's a great write-up about it [here](https://www.omgubuntu.co.uk/2023/04/pip-install-error-externally-managed-environment-fix). Anything >= 23.10 will break with just doing "pip install" I think

scriptlets/_put
plars: I like the idea of using `:` instead. That seems a little more intuitive

scriptlets/install_checkbox_agent_source
plars: that's true, I don't think we actually handle this yet. It does work ok for now, so we could defer it until it's an immediate problem. But we'll almost certainly have to resolve it at some point in the future.

install_tools.sh
boukeas: Thanks Paul, that's very useful. I will address this.

boukeas: @plars I have confirmed this will be a problem in future Ubuntu versions and researched possible solutions. My suggestion would be to use `pipx`, as long as we have it installed previously. The change to the existing code is minimal and it seems to fit our use-case perfectly ("install packages with cli entrypoints"). How would you feel about that?

scriptlets/install_checkbox_agent_source
boukeas: Using `pipx` (see comment below), this would be patched as below:
```
- (cd checkbox/checkbox-ng; sudo python3 setup.py install > /dev/null)
+ pipx install checkbox/checkbox-ng > /dev/null
```
I have tested this successfully in a multipass container.

scriptlets/_put
boukeas: @plars `_put` now uses the colon to indicate the target and a single colon means the home directory of `DEVICE_USER`. This mirrors how `scp` and similar programs behave.

scriptlets/install_checkbox_agent_source
boukeas: @plars I will install `pipx` explicitly and use it instead of `pip` here but do let me know if you think it might make sense to actually pre-install it when creating the container.

plars: It's even fine if we do both. Having it here is nice because we don't assume it's there and we ensure the requirements are installed. Having it in the pre-install means that installing it here will usually take no extra time.

install_tools.sh
plars: +1

scriptlets/install_checkbox_agent_source
boukeas: @plars `pipx` us not installed on the agent and used for installing `checkout_to_version`.

boukeas:
> Also I'm curious what the reason was for removing the bats tests. You said that it's probably less than adequate (I think you're right, but that's just one way that shell scripts can be very limiting). But since we don't really have a better option right now, is it better to leave it in until we do?

@plars I don't believe the existing tests are testing anything at all because none of them are actually running anything on a real system -- all system output is mocked. And that is just for the simple, low-level scriptlets. It would take an insane amount of effort to write BATS tests for the complex, higher-level scriptlets and they would still test nothing at all. In my opinion, [a comprehensive set of integration tests like these](https://github.com/canonical/hwcert-jenkins-jobs/pull/495#issuecomment-2273075191), that actually test the scriptlets within a realistic setting, is much more irrefutable proof that they work and are much more likely to catch regressions in the future.

jocave (approved):
Thanks George, I have no problems with this and given it should not impact any existing jobs we should be good to merge

PR #38, opened 2024-07-29
Remove Mantic as it no longer supported

Hook25:
We can't build for Mantic anymore so the canary pipeline will never start with this constraint. This removes the Mantic check.


fernando79513 (approved):
LGTM +1!

PR #40, opened 2024-08-27
fixed bug with snap info version (BugFix)

fernando79513:
This new version should work well even when using the default `distutils.version import LooseVersion` with python 3.8.

### DIFF
--- a/version-published/snap_info_utility.py
+++ b/version-published/snap_info_utility.py
@@ -65,6 +65,7 @@ def get_version_and_offset(version_str: str):
     # Try to parse the version and dev number
     try:
         Version(base_version)
+        dev_number = int(dev_number)
     except ValueError:
         raise SystemExit(f"Invalid version format: {version_str}")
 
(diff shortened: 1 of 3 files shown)


version-published/snap_info_utility.py
```diff
@@ -78,14 +79,16 @@ def get_previous_tag(base_version: str, repo_path: str):
...
+    tags = [t[1:] for t in tags if t.startswith("v")]
 
     # Get the previous tag corresponding to the base version. We have to do it
     # this way because the tags are only created once the version is published.
     # For example, 4.0.0.dev333 will use the previous tag v3.3.0 to calculate
     # the offset, not v4.0.0. The versions after 4.0.0 will use v4.0.0.
     try:
-        return next(t for t in tags if Version(t) < Version(base_version))
+        return next(
+            f"v{t}" for t in tags if Version(t) < Version(base_version)
```
boukeas, line 90: I find the handling of the `v` prefix a little hard to follow. The `v` is removed when the list of tags is filtered and then re-introduced in this line. Why is that better than e.g. `next(t[1:] for t in tags if Version(t) < Version(base_version))`

boukeas:
I think the `get_previous_version` function might be easier to test if it accepted a list of versions/tags as its input. The way it is right now, it relies on whatever `git tag` returns.

version-published/snap_info_utility.py
```diff
@@ -65,6 +65,7 @@ def get_version_and_offset(version_str: str):
     # Try to parse the version and dev number
     try:
         Version(base_version)
+        dev_number = int(dev_number)
```
boukeas, line 68: There's some repetition here, the function itself returns `int(dev_number)`.

version-published/snap_info_utility.py
fernando79513: It's basically the same. The `pakaging.version` module handled both versions with v or without v in front the same way, but `distutils.version` does not handle that.

I decided to create the list of tags without the trailing "v" because in the output of `get_version_and_offset` returns the base version without the "v", so I the `tags` list with the same format.

If you think it is more confusing that way, I can trim the starting "v" in the iterator.

version-published/snap_info_utility.py
fernando79513: Yes, I moved the `int()` inside the try so it fails beforehand if the dev number is not an integer. I will remove the second one

boukeas (approved):
Thanks for addressing the comments Fernando. All good.

PR #41, opened 2024-08-28
Add execution permissions to `add_private_ppa.py`

boukeas:
Whenever `add_private_ppa.py` is used in a certification job (e.g. Realtime, ESM of FIPS jobs) its use is preceded by a `chmod +x add_private_ppa.py` line.

This small PR adds the `u+x` permission to the file so that this `chmod` line can be removed from certification jobs.


omar-selo (approved):
Looks good

PR #42, opened 2024-09-04
[CERTTF-358] Amend and extend scriptlets for snap-testing jobs

boukeas:
# Description

[CERTTF-358](https://warthogs.atlassian.net/browse/CERTTF-358) aims to amend and extend the scriptlets in this repository so that they can be used for snap-testing jobs. It builds on previous work in the context of [CERTTF-337](https://warthogs.atlassian.net/browse/CERTTF-337), which focused on SRU jobs.

Please refer to the [companion PR](https://github.com/canonical/hwcert-jenkins-jobs/pull/569) in the `hwcert-jenkins-job` repo to see how these scriptlets are used in practice in the context of snap-testing jobs.

# Changelog

- A `_get` scriptlet is introduced to copy files from the device to the agent. Its functionality mirrors the `_put` scriptlet that performs the opposite function.
- `check_for_snap_changes` performs broader checks on the output of `snap changes` to determine if there are ongoing _or pending_ tasks (i.e. it doesn't only scan for `Doing`).
- `check_for_snap_changes` used to display the complete list of snap changes, when there are still ongoing changes. This may have been quite a long list, including snap changes from recent unrelated runs. This behaviour has now changed: now it only displays what the ongoing or pending snaps tasks are (if any).
- Tracing (`set -x`) is disabled within the `install_tools.sh` installer. This is because tracing generated substantial output that was unrelated to the test at hand and quite distracting for anyone reviewing the log. Tracing is restored to its previous setting at the end of the installer run. There is also some minor refactoring in the installer.
- A previous requirement for `_put`'s target to be prefixed by a `:` has been dropped. A `:` may still be used but this is now optional. This is for backwards-compatibility with scripts that might switch to using scriptlets. It is also because the same requirement in `_get` for the source(s) to be prefixed by a `:` would have looked awkward.
- A small modification was required in `install_psmisc`. The `psmisc` package is necessary because it includes `fuser`, which in turn is used in order to check that package operations are not ongoing. It is redundant in Core images, where the use of `apt-get` is not possible, thus making package operations not possible. The modification in `install_psmisc` is a refinement to the check for whether or not the installation of `psmisc` is required.

# Testing

See the [companion PR](https://github.com/canonical/hwcert-jenkins-jobs/pull/569) which includes a large set of snap-testing runs that make use of the scriptlets in this repo. Please do not approve this PR until reviewing those runs.

[CERTTF-358]: https://warthogs.atlassian.net/browse/CERTTF-358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CERTTF-337]: https://warthogs.atlassian.net/browse/CERTTF-337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scriptlets/check_for_snap_changes
```diff
@@ -37,20 +37,21 @@ all_complete() {
...
 
     RESULT=$?
     if [ "$RESULT" -eq 0 ]; then
-        # the change is still ongoing: display output as a diagnostic and return 1
-        cat "$OUTPUT"
+        # changes are still ongoing or pending: display output as a diagnostic and return 1
+        cat "$OUTPUT" | grep -E "\b(Doing|Undoing|Wait|Do|Undo)\b"
         rm "$OUTPUT"
         return 1
     else
```
plars, line 52: Completely optional, but it looks like you can drop the else block completely. If RESULT was 0 then it already returned, so we can just drop out of the if and handle the other branch here without an "else"

install_tools.sh
```diff
@@ -55,33 +27,73 @@ clone() {
...
+
+# assign default values if command-line arguments have not been provided
+TOOLS_PATH=${TOOLS_PATH:-$TOOLS_PATH_DEFAULT}
+BRANCH=${BRANCH:-$BRANCH_DEFAULT}
+
+# retrieve the tools from the repository
 fetch || (rm -rf $TOOLS_PATH && clone)
 
-# add scriptlets to agent's PATH
+# this is where the scriptlets are
```
plars, line 79: I don't think you actually *need* this comment, but it's ok.

plars (approved):
Looks good! +1 as it is, but I had 2 extremely minor, optional things below. Neither should block this from landing if you want to go ahead and merge it though.

PR #43, opened 2024-09-24
add(scriplets): run command on host via SSH

p-gentili:
Similarly to the `_run` scriplet, this PR introduces `_run`, to run commands over SSH on a target host. Username/Password are static and defined by the host image, SSH authentication is achieved via `sshpass`.

Tested it locally running a command on a host in the lab.

### DIFF
--- a/install_tools.sh
+++ b/install_tools.sh
@@ -88,10 +88,10 @@ wait_for_ssh --allow-degraded \
 && install_on_device \
 || exit 1
 
-# install agent dependencies
+echo "Installing agent dependencies"
 install_packages pipx python3-venv sshpass > /dev/null
 
-# install launcher tool on the agent
+echo "Installing agent tools"
 pipx install --spec $TOOLS_PATH/cert-tools/launcher launcher > /dev/null
 
 # restore tracing (if previously enabled)
--- a/install_tools.sh
+++ b/install_tools.sh
@@ -88,9 +88,10 @@ wait_for_ssh --allow-degraded \
 && install_on_device \
 || exit 1
 
+# install agent dependencies
+install_packages pipx python3-venv sshpass > /dev/null
+
 # install launcher tool on the agent
-echo "Installing pipx"
-install_packages pipx python3-venv > /dev/null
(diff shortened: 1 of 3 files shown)


boukeas:
@p-gentili just to be even safer, I would suggest creating a Testflinger job where the `test_cmds` are more or less like these:

```
# retrieve the tools installer
curl -Ls -o install_tools.sh https://raw.githubusercontent.com/canonical/hwcert-jenkins-tools/main/install_tools.sh
# install the scriptlets and other tools on the agent and the device, as necessary
export TOOLS_PATH=tools
source install_tools.sh \$TOOLS_PATH --branch <test-branch>

# ensure device is available before continuing
wait_for_ssh --allow-degraded

### use <keyword> here
```

Note how the installer retrieves the <test-branch> branch of the tools repo so you can test your latest and greatest.

You can submit the job and post the results here, in the description, to showcase that/how it works.

boukeas:
Also, and forgive my ignorance, is there any need for something like <wait-keyword> before running commands on it or are we assuming it is always responsive?

p-gentili:
Here's a test job, thanks for the example: https://testflinger.canonical.com/jobs/7d3bedfa-a277-4139-b0a8-11d14a359025 

Note that I had to manually install `sshpass`. Please let me know if you want to include the dependency in the install script instead.

> Also, and forgive my ignorance, is there any need for something like <wait-keyword> before running commands on it or are we assuming it is always responsive?

It's expected to be responsive in terms of SSH (we don't reboot it) but there might be something to wait for after snap refresh. I'll have to play a little bit more with it and, if needed, add a dedicated API on the control host.

boukeas:
> Note that I had to manually install sshpass. Please let me know if you want to include the dependency in the install script instead.

Good call. The installer script is supposed to take care of all scriptlet dependencies so introducing <test-keyword> means we need to add `sshpass` to the packages that the script installs. By the way, there's an `install_packages` scriptlet for that which you can also use on the device.

I've made that change now and here's a [run](https://testflinger.canonical.com/jobs/8658c03e-4cb9-4ec7-a6a5-8ef603996b08) that tests it.

boukeas (approved):

PR #44, opened 2024-09-27
Add: a scriptlet to refresh the machine snap and firmware

p-gentili:
This PR adds a scriptlet to refresh the snap when it's part of the DUT setup, upgrade firmware and wait until the service is stable. The need of such scriplet comes from the fact that we're going to start using multiple channels for the snap: `edge` for daily builds and canary validation and `beta` for any other usage in the lab.

Also, this would be the trigger for firmware upgrade which is not performed in other automatic way ATM.

### Tests
 Snap up-to-date: https://testflinger.canonical.com/jobs/b88002a2-8c76-4231-a695-c7d60d298766
 Snap outdated: https://testflinger.canonical.com/jobs/957c6b48-d212-49ad-b5e8-29b66bd86267
 FW outdated: https://testflinger.canonical.com/jobs/c339d95f-397f-4277-aa1f-1e7feb4b26ec
 To edge channel: https://testflinger.canonical.com/jobs/143a159c-1dbd-4ee4-8673-01b6c1ce7822

### DIFF
--- a/scriptlets/refresh_zapper_if_needed
+++ b/scriptlets/refresh_zapper_if_needed
@@ -1,0 +1,74 @@
+#!/usr/bin/env bash
+
+# Update the Zapper snap (and firmware) if needed
+#
+# Description:
+#
+# If a Zapper is part of the setup, refresh the snap
+# to the target channel (default is `beta`), upgrade firmware
+# and wait until the service is stable. 
+
+usage() {
+    echo "Usage: $(basename ${BASH_SOURCE[0]}) [--channel]"
+}
+
+check_for_zapper () {
+    (
+        trap EXIT
+        source "$(dirname ${BASH_SOURCE[0]})/defs/check_for_zapper_ip" > /dev/null
+    )
+}
+
+wait_zapper_service_stable() {
+    previous_addon_count=-1
+    stable_count=0
+    max_stable_checks=3
+    while true; do
+        current_addon_count=$(eval "_run_zapper zapper addon list 2> /dev/null" | wc -l)
+        if [[ "$current_addon_count" -eq "$previous_addon_count" ]]; then
+          ((stable_count++))
+        else
+          stable_count=0
+        fi
+
+        if [[ "$stable_count" -ge "$max_stable_checks" ]]; then
+          break
+        fi
+
+        previous_addon_count="$current_addon_count"
+    done
+}
+
+CHANNEL=beta
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --channel)
+            CHANNEL=$2
+            shift
+            ;;
+        *)
+            usage
+            echo "Error: Invalid argument $1"
+            exit 1
+            ;;
+    esac
+    shift
+done
+
+check_for_zapper
+if [ $? -ne 0 ]; then
+    exit 0
+fi
+
+message=$(_run_zapper sudo snap refresh zapper --amend --channel $CHANNEL 2>&1)
+if echo "$message" | grep -q 'no updates available'; then
+    echo "Snap was up-to-date, restarting the service instead..."
+    _run_zapper sudo snap restart zapper
+fi
+wait_zapper_service_stable
+
+message=$(_run_zapper zapper firmware update -y --allow-older)
+if echo "$message" | grep -q 'update in progress'; then
+    echo "Firmware(s) got upgraded, waiting again for the service to stabilize..."
+    wait_zapper_service_stable
+fi
--- a/scriptlets/refresh_zapper_if_needed
+++ b/scriptlets/refresh_zapper_if_needed
@@ -28,8 +28,7 @@ while [[ "$#" -gt 0 ]]; do
     shift
 done
 
-# Running in a subshell to not exit with error like it normally would when failing
-(source "$(dirname ${BASH_SOURCE[0]})/defs/check_for_zapper_ip" >> /dev/null)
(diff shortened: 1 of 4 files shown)


scriptlets/refresh_zapper_if_needed
```diff
@@ -0,0 +1,74 @@
...
+        *)
+            usage
+            echo "Error: Invalid argument $1"
+            exit 1
+            ;;
+    esac
+    shift
+done
+
+check_for_zapper
```
boukeas: You could replace this with `check_for_ || exit 0`

scriptlets/refresh_zapper_if_needed
```diff
@@ -0,0 +1,74 @@
...
+        source "$(dirname ${BASH_SOURCE[0]})/defs/check_for_zapper_ip" > /dev/null
+    )
+}
+
+wait_zapper_service_stable() {
+    previous_addon_count=-1
+    stable_count=0
+    max_stable_checks=3
+    while true; do
+        current_addon_count=$(eval "_run_zapper zapper addon list 2> /dev/null" | wc -l)
```
boukeas: I hope I am not being thick here but would it be a problem to use:
```
current_addon_count=$(<run-keyword> addon list 2> /dev/null | wc -l)
```

scriptlets/refresh_zapper_if_needed
```diff
@@ -0,0 +1,74 @@
...
+}
+
+check_for_zapper () {
+    (
+        trap EXIT
+        source "$(dirname ${BASH_SOURCE[0]})/defs/check_for_zapper_ip" > /dev/null
+    )
+}
+
+wait_zapper_service_stable() {
```
boukeas: Would you consider turning this into a separate `wait_for_service_stable` scriptlet? (especially if you think it may be reused in other scripts as well?

boukeas: Perhaps it would be worth adding a comment here so that anyone reading this doesn't have to spend any time understanding the logic. Something like "Wait until the number of addons is stable for max_stable_checks consecutive times"

scriptlets/refresh_zapper_if_needed
```diff
@@ -0,0 +1,74 @@
...
+        else
+          stable_count=0
+        fi
+
+        if [[ "$stable_count" -ge "$max_stable_checks" ]]; then
+          break
+        fi
+
+        previous_addon_count="$current_addon_count"
+    done
```
boukeas, line 39: Don't you need some sort of sleep here between consecutive polls?

scriptlets/refresh_zapper_if_needed
```diff
@@ -0,0 +1,74 @@
...
+    esac
+    shift
+done
+
+check_for_zapper
+if [ $? -ne 0 ]; then
+    exit 0
+fi
+
+message=$(_run_zapper sudo snap refresh zapper --amend --channel $CHANNEL 2>&1)
```
boukeas, line 63: Any chance this might trigger a reboot? Any chance other snap changes or auto-refreshes can be ongoing and/or trigger a reboot?

scriptlets/refresh_zapper_if_needed
p-gentili: it's now described in its own "docstring"

scriptlets/refresh_zapper_if_needed
p-gentili: no because the `addon list` takes seconds to run, and already gives the right pace

scriptlets/refresh_zapper_if_needed
p-gentili: not at all, we control that snap and for now there's no reboot action on snap refresh or configure.

scriptlets/refresh_zapper_if_needed
```diff
@@ -0,0 +1,45 @@
...
+            usage
+            echo "Error: Invalid argument $1"
+            exit 1
+            ;;
+    esac
+    shift
+done
+
+# Running in a subshell to not exit with error like it normally would when failing
+(source "$(dirname ${BASH_SOURCE[0]})/defs/check_for_zapper_ip" >> /dev/null)
```
boukeas: I may be missing something but would this work the way you intend it to? If so, I would personally prefer it because it makes the fact that it will exit (if the first command fails) explicit.
```
source "$(dirname ${BASH_SOURCE[0]})/defs/check_for_ip" >> /dev/null || exit 0
```

boukeas (approved):

scriptlets/refresh_zapper_if_needed
p-gentili: Unfortunately that still exits with `RC=1` because you're sourcing the file, making it part of the main script. If we make that `check_for_ip` executable instead, we can make it work like this:
```
$(dirname ${BASH_SOURCE[0]})/defs/check_for_ip   >> /dev/null || exit 0
```
Let me know if it's ok for you George to make it executable.

boukeas: Yes, it's definitely OK to make it executable and, in that case, move it out of `defs` (i.e. directly under scriptlets) so that you can run it directly: `check_for_ip >> /dev/null || exit 0`. Anything in `defs` is meant to be sourced or executed on the device over ssh (i.e. one-off, without copying it over) and, to be perfectly honest, I don't remember/understand why `check_for_device_ip` is sourced in the other scripts instead of being an executable as well.

PR #45, opened 2024-09-27
New: Scriplet that clean everything from a previous run 

Hook25:
We have quite a few lines in all templates dedicated to clean up what the previous job left behind. These are mainly used for noprovision machine. Given that I need something very similar, this is my implementation as a scriplet we can call instead of the multiple lines.

Tests:
- Missing `--im-sure` run (which I definetly intended to create and didn't do it by mistake): https://github.com/canonical/checkbox/actions/runs/11103858615/job/30846714818
- Actual run: https://github.com/canonical/checkbox/actions/runs/11104217105/job/30847831704#step:5:512

### DIFF
--- a/install_tools.sh
+++ b/install_tools.sh
@@ -35,7 +35,7 @@ install_on_device() {
 
     # copy selected scriptlets over to the device
     # and then move them somewhere in the device's PATH
-    DEVICE_SCRIPTLETS=(retry check_for_packages_complete wait_for_packages_complete install_packages)
+    DEVICE_SCRIPTLETS=(retry check_for_packages_complete wait_for_packages_complete install_packages clean_machine)
     _put "${DEVICE_SCRIPTLETS[@]/#/$SCRIPTLETS_PATH/}" :
     _run sudo mv "${DEVICE_SCRIPTLETS[@]}" $REMOTE_PATH
 
--- a/scriptlets/clean_machine
+++ b/scriptlets/clean_machine
@@ -38,7 +38,7 @@ if [ -n "$CHECKBOX_DEBS" ]; then
     echo "Found the following Checkbox debs"
     echo "  " $CHECKBOX_DEBS
     echo "Deleting them..."
-    apt purge -y -qq $CHECKBOX_DEBS
+    apt-get purge -y -qq $CHECKBOX_DEBS
 fi
 
 # delete old sessions, as Checkbox will try to resume them when started,
@@ -58,11 +58,10 @@ fi
 # from the wrong source. Also removing /ppa and /testing which are legacy names
 # of /edge and /beta that may still be used somewhere
 echo "Trying to remove all Checkbox ppas"
-add-apt-repository --remove -y ppa:checkbox-dev/ppa
-add-apt-repository --remove -y ppa:checkbox-dev/testing
-add-apt-repository --remove -y ppa:checkbox-dev/edge
-add-apt-repository --remove -y ppa:checkbox-dev/beta
-add-apt-repository --remove -y ppa:checkbox-dev/stable
+RISKS="ppa testing edge beta stable"
+for risk in $RISKS; do
+  sudo add-apt-repository --remove -y ppa:checkbox-dev/$risk > /dev/null && echo "Removed ppa:checkbox-dev/$risk"
+done
 
 # Also remove any cloned version of Checkbox in any home, as this will clash
 # with anything trying to provision from source
--- a/scriptlets/clean_machine
+++ b/scriptlets/clean_machine
@@ -49,7 +49,7 @@ if [ -d "/var/tmp/checkbox-ng" ]; then
 fi
 
 # delete any provider that was globally "developed" (installed in editable mode)
-if [ -d "/var/tmp/checkbox-providers-develop/"]; then
+if [ -d "/var/tmp/checkbox-providers-develop/" ]; then
   echo "Removing all developed providers"
   rm -rf /var/tmp/checkbox-providers-develop
 fi
--- a/scriptlets/clean_machine
+++ b/scriptlets/clean_machine
@@ -1,5 +1,5 @@
 #!/bin/bash
-# this is a generic clean script that deletes any thing that may break the
+# This is a generic clean script that deletes any thing that may break the
 # current test run due to a previous run. This is especially useful for
 # noprovision machines.
 
--- a/scriptlets/clean_machine
+++ b/scriptlets/clean_machine
@@ -23,7 +23,7 @@ set +e
 # delete all checkbox snaps, this includes any frontend or custom frontend
 # (they all contain checkbox in their name). This should also purge any
 # running checkbox*agent service
-CHECKBOX_SNAPS=$(snap list | \grep -E 'checkbox' | \grep -P -o "^[\w-]+")
+CHECKBOX_SNAPS=$(snap list | grep -E 'checkbox' | grep -P -o "^[\w-]+")
 if [ -n "$CHECKBOX_SNAPS" ]; then
     echo "Found the following Checkbox snaps"
     echo "  " $CHECKBOX_SNAPS
@@ -33,7 +33,7 @@ fi
 
 # delete all installed Checkbox debian packages. This should also purge any
 # previously installed provider
-CHECKBOX_DEBS=$(apt list --installed | \grep checkbox | \grep -P -o "^[\w-]+")
+CHECKBOX_DEBS=$(apt list --installed | grep checkbox | grep -P -o "^[\w-]+")
 if [ -n "$CHECKBOX_DEBS" ]; then
     echo "Found the following Checkbox debs"
     echo "  " $CHECKBOX_DEBS
--- a/scriptlets/clean_machine
+++ b/scriptlets/clean_machine
@@ -28,7 +28,7 @@ if [ -n "$CHECKBOX_SNAPS" ]; then
     echo "Found the following Checkbox snaps"
     echo "  " $CHECKBOX_SNAPS
     echo "Deleting them..."
-    snap remove $CHECKBOX_SNAPS
+    sudo snap remove $CHECKBOX_SNAPS
 fi
 
 # delete all installed Checkbox debian packages. This should also purge any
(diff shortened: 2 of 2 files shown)


scriptlets/clean_machine
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+# this is a generic clean script that deletes any thing that may break the
```
boukeas, line 2: Typo "This"

scriptlets/clean_machine
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
```
boukeas, line 1: This is supposed to run on the DUT, right? I think this is perhaps something that needs to be explicit in the "docstring". Also, if it's supposed to run on the device then it needs to be one of the scriptlets that is copied over to the device in the [installer script](https://github.com/canonical/hwcert-jenkins-tools/blob/fca2f9f2269dcd0f6120feaf1c75c157854accc8/install_tools.sh#L38).

Just to note (and I am really not sure of the answer myself): so far we have been running this sort of thing from the agent using individual `_run` commands. This shifts the approach to running the commands/script directly on the device. I am not sure which is best but it does need to be weighed and considered.

scriptlets/clean_machine
```diff
@@ -0,0 +1,73 @@
...
+
+# delete all checkbox snaps, this includes any frontend or custom frontend
+# (they all contain checkbox in their name). This should also purge any
+# running checkbox*agent service
+CHECKBOX_SNAPS=$(snap list | \grep -E 'checkbox' | \grep -P -o "^[\w-]+")
+if [ -n "$CHECKBOX_SNAPS" ]; then
+    echo "Found the following Checkbox snaps"
+    echo "  " $CHECKBOX_SNAPS
+    echo "Deleting them..."
+    snap remove $CHECKBOX_SNAPS
```
boukeas, line 31: Don't you need a `sudo` here?

scriptlets/clean_machine
```diff
@@ -0,0 +1,73 @@
...
+  echo "if you are sure launch: `basename "$0"` --im-sure"
+  exit 1
+fi
+
+set +e
+
+# delete all checkbox snaps, this includes any frontend or custom frontend
+# (they all contain checkbox in their name). This should also purge any
+# running checkbox*agent service
+CHECKBOX_SNAPS=$(snap list | \grep -E 'checkbox' | \grep -P -o "^[\w-]+")
```
boukeas, line 26: Enlighten me: why `\grep` and not plain `grep`?

scriptlets/clean_machine
```diff
@@ -0,0 +1,73 @@
...
+fi
+
+# delete all installed Checkbox debian packages. This should also purge any
+# previously installed provider
+CHECKBOX_DEBS=$(apt list --installed | \grep checkbox | \grep -P -o "^[\w-]+")
+if [ -n "$CHECKBOX_DEBS" ]; then
+    echo "Found the following Checkbox debs"
+    echo "  " $CHECKBOX_DEBS
+    echo "Deleting them..."
+    apt purge -y -qq $CHECKBOX_DEBS
```
boukeas, line 41: Don't you need a `sudo` here?

scriptlets/clean_machine
```diff
@@ -0,0 +1,73 @@
...
+
+# delete any provider that was globally "developed" (installed in editable mode)
+if [ -d "/var/tmp/checkbox-providers-develop/"]; then
+  echo "Removing all developed providers"
+  rm -rf /var/tmp/checkbox-providers-develop
+fi
+
+# remove all ppa repos pre-configured as they may make the installer install
+# from the wrong source
+# leave also /ppa and /testing, legacy names of /edge and /beta
```
boukeas, line 59: The comment may be a little misleading here. One could it read it as "remove all these PPAs but _leave_ edge and beta" but it's obviously not what you mean.

scriptlets/clean_machine
```diff
@@ -0,0 +1,73 @@
...
+echo "Trying to remove all Checkbox ppas"
+add-apt-repository --remove -y ppa:checkbox-dev/ppa
```
boukeas, line 61: Don't you need `sudo` here?

boukeas, line 61: I seem to remember this part being a little "noisy" in the SRU runs when the PPAs have not been previously added to the system. Would it perhaps make sense to suppress the output (e.g. redirect to `/dev/null`) and only print some sort of message when the command is successful, for example:
```
sudo add-apt-repository --remove -y ppa:checkbox-dev/$RISK > /dev/null && echo "Removed ppa:checkbox-dev/$RISK"
```
Also, would you consider using a for-loop over $RISK here?

scriptlets/clean_machine
boukeas, line 41: Does using `apt` instead of `apt-get` still generate a warning that `apt` shouldn't be used in scripts because it's interface is not stable? If so, perhaps best to go for `apt-get`

scriptlets/clean_machine
```diff
@@ -0,0 +1,73 @@
...
+add-apt-repository --remove -y ppa:checkbox-dev/testing
+add-apt-repository --remove -y ppa:checkbox-dev/edge
+add-apt-repository --remove -y ppa:checkbox-dev/beta
+add-apt-repository --remove -y ppa:checkbox-dev/stable
+
+# Also remove any cloned version of Checkbox in any home, as this will clash
+# with anything trying to provision from source
+CHECKBOX_SOURCE_CLONES=$(find /home/ -maxdepth 2 -name "*checkbox*" -type d)
```
boukeas, line 69: Just a suggestion but you can delete the directories you find directly within `find`
```
echo "Removing all clones of Checkbox on the machine"
find /home/ -maxdepth 2 -name "*checkbox*" -type d -exec rm -rf {} +
```
If I remember correctly this should also display the ones that are found and removed.

boukeas (changes requested):
This is really useful and should clean up the SRU script considerably. I have made a few comments and suggestions. In the end, can you please also include a link to one or two Testflinger jobs where the scriptlet is used, as a demonstration of how it works?

scriptlets/clean_machine
Hook25: I think this is a bit easier to read as is, as basically every  command in this would use `_run` and I can see only downsides to it. The only thing we can get out of `_run`ing every command in this via individual `_run` is network latency, or am I missing something?

scriptlets/clean_machine
Hook25: I have a lot of aliases on my machine, I will fix this, I was copy pasting from my console

scriptlets/clean_machine
Hook25: ditto above

scriptlets/clean_machine
Hook25: ditto above

Hook25: Makes sense, originally I wanted to detect installed ppas and remove them, but that is harder than just removing them all

scriptlets/clean_machine
Hook25: I will fix it, same issue as in `\grep`

scriptlets/clean_machine
Hook25: yes, I wanted for "simmetry" to only print the "Removing all clones..." message when there is actually something to remove, but I can change this.

scriptlets/clean_machine
Hook25: Also, I expected the installer script to copy everything in the scriplet dir with a glob, any reason we are maintaining an explicit list over there?

scriptlets/clean_machine
boukeas: Yes, I don't think it's good practice to do that. Use `sudo` only when you need to.

scriptlets/clean_machine
Hook25: I'm keeping one `apt` command, `apt list` as I don't know if there is a "clean" alternative, I think to not use it one needs to fetch the installed packages via dpkg, I would rather not

scriptlets/clean_machine
Hook25: Fixed beside 1

scriptlets/clean_machine
Hook25: Fixed

scriptlets/clean_machine
Hook25: Fixed

scriptlets/clean_machine
Hook25: Fixed

scriptlets/clean_machine
Hook25: Fixed

scriptlets/clean_machine
Hook25: Fixed

Hook25: Fixed

scriptlets/clean_machine
boukeas: Just to clarify: you decided against using the `-exec` version?

install_tools.sh
```diff
@@ -35,7 +35,7 @@ install_on_device() {
 
     # copy selected scriptlets over to the device
     # and then move them somewhere in the device's PATH
-    DEVICE_SCRIPTLETS=(retry check_for_packages_complete wait_for_packages_complete install_packages)
+    DEVICE_SCRIPTLETS=(retry check_for_packages_complete wait_for_packages_complete install_packages clean_machine)
```
boukeas, line 38: Looking at the name of the script, can I suggest renaming it to something that suggests it cleans traces of Checkbox, rather than cleaning a machine generally?

Hook25: I expect this script to be updated to be the one stop shop for cleaning a noprovision machine. If any command to clean up a machine has to be added, it must be added here. I only wrote the Checkbox side of business because that is what I know best

boukeas (approved):

scriptlets/clean_machine
Hook25: yes as I think the current one produces a better output log

Hook25:
Merged as workflow did pass here: https://github.com/canonical/checkbox/actions/runs/11104965148/job/30850115894

Failure is due to network fails

PR #46, opened 2024-10-01
feat(scriptlets): introduce `git_get`

boukeas:
# Description

There are instances where our certification scripts need to fetch a specific commit or branch from a repo and there is little point to performing a full clone when a light, shallow retrieval would suffice. Apart from saving time, network resources and space, there are indications that this is actually _required_ in cases where the device running the clone is resource-constrained.

This PR introduces the `git_get` scriptlet to that end (see comments for a test run).


Hook25:
Tested this with the new Checkbox action here: https://github.com/canonical/checkbox/actions/runs/11141378594/job/30962120438

It works.

scriptlets/git_get
```diff
@@ -0,0 +1,82 @@
...
+            fi
+            ;;
+        *)
+            echo "Error: Invalid argument $1";
+            usage
+            exit 1
+            ;;
+    esac
+    shift
+done
```
Hook25, line 64: imo this is flying to close to the sun as far as maximum complexity I would like to see in a bash script.  Still, if you want to keep it in bash consider using `getopts` and please fix the fact that I can provide both a commit and a branch to this it will not complain!

scriptlets/git_get
Hook25, line 1: branch is called git_get, should we call it git_shallow_get or something less ugly but to the same meaning? If one does actually need the history, this not the script for them and the name doesn't communicate it well enough imo

Hook25 (approved):
Consider the following comments, its ok to land it as is imo

scriptlets/git_get
boukeas: Done, renamed to `git_get_shallow`.

scriptlets/git_get
boukeas: That's a fair point regarding length but most of it is argument parsing, so hopefully not too complex. Using `getopts` is a great suggestion but in this particular case I don't think it would really help as there are no short-option arguments. Your comment did get me thinking though and I have slightly refactored to hopefully make the argument parsing clearer (i.e. to better reflect the existence of two different "modes".

Hook25:
green pipeline using this new version of the script

PR #47, opened 2024-10-03
fix: `_run` injects the device scriptlet location in the device PATH

boukeas:
# Description

Previously, `defs/scriptlet_path` was used to scan the device for an appropriate location to copy over selected scriptlets. That location needed to be already in the device's `PATH`, so that the scriptlets could be easily invoked. However, this presented problems in certain Core systems -- typically `/snap/bin` would be used to copy the scriptlets but this sometimes failed.

This PR does away with `defs/scriptlet_path`. The scriptlets are now copied to a pre-specified location on the device (the `TOOLS_PATH_DEVICE` directory on the default user's home folder) and the `_run` scriptlet _injects_ that location into the devices' `PATH`. 

# Tests

The issue was reported [here](https://chat.canonical.com/canonical/pl/ojm3g99tofn39fs78o7pgz86xy). The original test job was modified to retrieve the installer and scriptlets from this branch, subsequently [completing the run](http://testflinger.canonical.com/jobs/96eb25b3-bc78-4d30-940e-d9395fb6fe57 successfully.

This [test run](http://10.102.156.15:8080/job/cert-cm3-arm64-pi-kernel-18-pi-beta/73/console) was also previously failing at the point where the installer move some of the scriptlets on the device:
```
13:36:02 Installing selected scriptlets on the device
13:36:13 2024-09-30 12:35:56,392 cm3001 WARNING: DEVICE CONNECTOR: Tests failed, rc=1
```
The job was modified to retrieve the installer and scriptlets from this branch, subsequently [completing the run](http://10.102.156.15:8080/job/cert-cm3-arm64-pi-kernel-18-pi-beta/74/) successfully.

Finally, an SRU job was also deployed from this branch, to make sure that this PR not only fixes the problem but also doesn't break existing jobs. SRU jobs make use of the scriptlets that are copied on the device, so they are a perfect test for whether or not the path injection works. The job run [completed successfully](http://10.102.156.15:8080/job/cert-stock-sru-focal-desktop-xps13-7390-c2/109/consoleFull).


Hook25 (approved):
+1, LGTM

PR #48, opened 2024-10-07
Install Checkbox scriplet

Hook25:
We often need to install the checkbox snap during jobs. We need a way to install a runtime and a frontend of some risk from some channel. Once that is done we fetch the installed version (to install checkbox in the agent as well) and check if the Checkbox service has started.

This scriplet merges most of these lines into one script that we can use to the same effect.

Used in all of these: https://github.com/canonical/checkbox/actions/runs/11292389194
Also here after refactoring: https://github.com/canonical/checkbox/actions/runs/11366354867


Hook25:
For the reviewer: do we want this to also provision checkbox from source on the DUT?

scriptlets/install_checkbox_snaps.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+# Installs Checkbox and exports the installed version on the DUT
```
boukeas, line 2: I would slightly amend that to "Installs the Checkbox snap on the device/DUT and exports the installed version to...".

boukeas:
> For the reviewer: do we want this to also provision checkbox from source on the DUT?

In SRU tests we do that on the agent, not the DUT. Is that what you meant? If so, it would be quite handy.

scriptlets/install_checkbox_snaps.sh
```diff
@@ -0,0 +1,46 @@
...
+echo "Installing Checkbox snaps"
+
+RISK=$1
+shift
+RUNTIME_NAME=$1
+shift
+FRONTEND_NAME=$1
+shift
+FRONTEND_TRACK=$1
+shift
```
boukeas, line 19: Are there reasonable default values for any of these variables? If so, I vote for making them optional. I know it will probably make this script longer in order to parse them but it will also make the calls to the script more explicit.

scriptlets/install_checkbox_snaps.sh
```diff
@@ -0,0 +1,46 @@
...
+
+wait_for_snap_complete
+_run sudo snap install $RUNTIME_NAME --channel=latest/$RISK
+wait_for_snap_complete
```
boukeas, line 23: The approach I've followed for every `snap-install` so far is to use it with `--no-wait`, in combination with `wait_for_snap_changes` (not `wait_for_snap_complete`). This makes it more robust and immune to any interference from auto-updates or reboots.

Hook25: (I wonder what is the difference between `wait_for_snap_changes` and `wait_for_snap_complete`, I used the other one in the beginning but by name this looked like it was more correct)

Hook25: (I wonder what is the difference between `wait_for_snap_changes` and `wait_for_snap_complete`, I used the other one in the beginning but by name this looked like it was more correct)

scriptlets/install_checkbox_snaps.sh
Hook25: No, I want all of these to be explicit, I expect this to be called like this
```
install_checkbox_snaps beta checkbox22 checkbox 22.04
```

scriptlets/install_checkbox_snaps.sh
boukeas: `wait_for_snap_complete` was the name of the inline function previously used in snap-testing jobs. There is also a `wait_for_snap_complete` scriptlet written some time ago that was a variation of that function but had never been used. I do not clearly recall why I chose to name the reinvented script differently -- I assume I wanted to reflect the `snap changes` command that it monitors. I acknowledge it is confusing but I claim to compensate for that by the fact that it seems to work very well :grin:

scriptlets/install_checkbox_snaps
```diff
@@ -0,0 +1,49 @@
...
+echo "Installing Checkbox snaps"
+
+RISK=$1
+shift
+RUNTIME_NAME=$1
+shift
+FRONTEND_NAME=$1
+shift
+FRONTEND_TRACK=$1
+shift
```
boukeas, line 19: How about removing these `shift`s and accessing `$1` to `$4` directly?

Hook25: makes sense, idk why I did it this way

scriptlets/install_checkbox_debs
```diff
@@ -0,0 +1,25 @@
...
+_run install_packages checkbox-ng python3-checkbox-ng checkbox-provider-base checkbox-provider-resource checkbox-provider-sru fswebcam obexftp wmctrl iperf mesa-utils vim pastebinit fwts xorg-dev gir1.2-clutter-1.0
+wait_for_packages_complete
+
+export CHECKBOX_VERSION=$(_run checkbox-cli --version)
+[ -z "$CHECKBOX_VERSION" ] && echo "Error: Unable to retrieve Checkbox version from device" && exit 1
+
+check_for_checkbox_service
+
+echo "Installing checkbox $CHECKBOX_VERSION on the agent container from source"
+install_checkbox_agent_source $CHECKBOX_VERSION
```
boukeas, line 25: I am wondering if it would make sense to incorporate these lines into `install_checkbox_agent_source`. They are almost identical for both the snap and deb version, except for the fact that the FRONTEND_NAME is fixed (not an input) for the deb version.

If we were to do that, we wouldn't need to duplicate the code here. We would just call `install_checkbox_agent_source`, just making sure that `FRONTEND_NAME` is set in both cases.

One argument I can think of _against_ this suggestion is if there are cases when we would need to use `install_checkbox_agent_source` with an arbitrary version, instead of one we retrieve from the device.

Hook25: my argument for why this should call the other script and not the other way around is that this script knows how to fetch the checkbox version while the other doesnt

boukeas (approved):
Stellar! +1

PR #50, opened 2024-10-10
Improve the reliability of Test Observer's rerunner script

omar-selo:
Resolves https://warthogs.atlassian.net/browse/RTW-374

Changes:
- Adds requests timeout to avoid hanging
- Adds requests retries to improve reliability
- Brings back CI that tests this script as it seems to have been accidentally deleted at some point

Testing:
- Automated testing
- Also manually deployed this branch [here](http://10.102.156.15:8080/job/test-executions-rerunner/44385/) and sure enough it reran the [job](http://10.102.156.15:8080/job/cert-rpi400-arm64-pi-kernel-18-pi-beta/106/) correctly

### DIFF
--- a/scriptlets/test-executions-rerunner/test_executions_rerunner.py
+++ b/scriptlets/test-executions-rerunner/test_executions_rerunner.py
@@ -23,7 +23,7 @@ logging.basicConfig(level=logging.INFO)
 requests = Session()
 retries = Retry(
     total=3,
-    backoff_factor=0.1,
+    backoff_factor=3,
     status_forcelist=[408, 429, 502, 503, 504],
     allowed_methods={"POST", "GET", "DELETE", "PUT"},
 )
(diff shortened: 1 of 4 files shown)


scriptlets/test-executions-rerunner/test_executions_rerunner.py
```diff
@@ -8,14 +8,28 @@
...
 from requests.auth import HTTPBasicAuth
+from urllib3.util import Retry
 
 logging.basicConfig(level=logging.INFO)
 
+
+requests = Session()
+retries = Retry(
+    total=3,
+    backoff_factor=0.1,
```
nancyc12, line 26: I'm wondering if we should set `backoff_factor` to a bigger number (for example `backoff_factor=3`, so we will have jitter as 5, 10, 20) to give it more chance to get through the gateway-related errors that might not be recovered shortly?

nancyc12 (approved):
Thanks for making this improvement! It looks good to me. I only have one small suggestion as below.

scriptlets/test-executions-rerunner/test_executions_rerunner.py
omar-selo: Good idea

PR #51, opened 2024-10-16
[CERTTF-425] feat: add scriptlets for checking and parsing the scenario file

boukeas:
# Description

This PR introduces scriptlets that check and parse a "scenario" file, a JSON file that holds all the information required for a specific test run. Specifically:
- `check_for_scenario_file` makes sure the `SCENARIO_FILE` environment variable has been set and points to a valid JSON file
- `scenario` accesses the scenario file and retrieves a specific key

# Tests

This [Testflinger job run](https://testflinger.canonical.com/jobs/2b022aa7-3aa2-4044-b5a4-78935e6e64ca) illustrates how these scriptlets would be used during testing, with the scenario file provided as an attachment.

### DIFF
--- a/scriptlets/check_for_scenario_file
+++ b/scriptlets/check_for_scenario_file
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+# Check if the environment variable $SCENARIO_FILE is set
+# and points to a valid JSON file
+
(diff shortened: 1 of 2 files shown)


Hook25 (approved):
+1, ty for clarifying that big sed and the comment, it is way easier to read now

PR #53, opened 2024-11-26
[RTW-399] Update select scriptlets for use in snap-testing jobs

boukeas:
This PR primarily modifies the following scriptlets:
- `install_checkbox_snaps`
  - The name of the Checkbox runtime snap is no longer provided as an argument to the script but rather determined by the script itself, using the `base` of the frontend snap as a heuristic and falling back on the Core version when that `base` is not available.
  - The frontend Checkbox snap is installed in `--devmode` by default, falling back to installing in `--classic` mode if an error occurs.
  - Some default plugs (such as `provider-certification-client`) are not available on all custom Checkbox versions, so the script now installs them conditionally, i.e. after checking that it should.
- `clean_machine`
    - The script determines if package handling is available on the device (e.g. it is not available on Core systems) and only performs package-handling operations conditionally.
    - Some additional cleaning up of providers is performed and all cleaning up is performed using `sudo`, as permission problems were encountered on some systems during testing.

The modifications were made in the context of [RTW-399](https://warthogs.atlassian.net/browse/RTW-399), so that these scriptlets could be used for the snap-testing Jenkins jobs. Additional information (and tests) can be found in the [companion PR](https://github.com/canonical/hwcert-jenkins-jobs/pull/618) in the jobs repo.

[RTW-399]: https://warthogs.atlassian.net/browse/RTW-399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scriptlets/install_checkbox_snaps
+++ b/scriptlets/install_checkbox_snaps
@@ -31,7 +31,7 @@ FRONTEND_TRACK=$3
 FRONTEND_CHANNEL="$FRONTEND_TRACK/$RISK"
 
 # the runtime name is "checkboxYY";
-# YY is extracted from the "base" field of the front-end snap, which should be coreYY;
+# YY is extracted from the "base" field of the frontend snap, which should be coreYY;
 # the final "grep ." will cause this script to exit if there's no match
 RUNTIME_NAME="checkbox$(_run "snap info --verbose $FRONTEND_NAME" | sed -n 's/^base:\s*core\([0-9]\{2\}\)$/\1/p' | grep .)"
 RUNTIME_CHANNEL="latest/$RISK"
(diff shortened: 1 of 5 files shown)


scriptlets/install_checkbox_snaps
```diff
@@ -10,77 +10,107 @@
...
+}
+
+get_runtime() {
+    # get the store token from the device, if available
+    STORE=$(_run "snap model --assertion" | sed -n 's/^store:\s\(.*\)$/\1/p')
+    # get the snap info from the snap store and extract the base
+    BASE=$(curl -s -H 'Snap-Device-Series: 16' ${STORE:+ -H "Snap-Device-Store: $STORE"} "https://api.snapcraft.io/v2/snaps/info/$FRONTEND_NAME?fields=base" \
+    | jq -r \
+    --arg FRONTEND_TRACK $FRONTEND_TRACK --arg RISK $RISK --arg ARCH $(get_ubuntu_arch) \
+    '.["channel-map"] | .[] | select(.channel.track == $FRONTEND_TRACK and .channel.risk == $RISK and .channel.architecture == $ARCH) | .base')
```
Hook25, line 33: I would prefer this wasn't so complicated to read, why don't we use a temp var and put some spacing to the jq query, I really can't read it like this. Does this look good to you?
```
jq -r  \
  --arg FRONTEND_TRACK $FRONTEND_TRACK \
  --arg RISK $RISK --arg ARCH $(get_ubuntu_arch) \
    '
      .["channel-map"] | .[] | 
      select(
          .channel.track == $FRONTEND_TRACK and 
          .channel.risk == $RISK and
          .channel.architecture == $ARCH
      ) | 
      .base')
```

scriptlets/install_checkbox_snaps
```diff
@@ -10,77 +10,107 @@
...
+    # use the base as a heuristic to derive the Checkbox runtime
+    if [ "$BASE" = "null" ]; then
+      # no base:" fall back to using the Core version
+      # (the final "grep ." will cause the command to fail if there's no match)
+      RUNTIME_NAME="checkbox$(_run "cat /etc/os-release" | sed -n 's/PRETTY_NAME="Ubuntu Core \([0-9]\{2\}\)"$/\1/p' | grep .)"
```
Hook25, line 38: Pretty sure (99%, we can confirm with store) that no base => core => checkbox16

scriptlets/install_checkbox_snaps
```diff
@@ -10,77 +10,107 @@
...
-  set -e
-
-  # force connect also the default plugs. This is necessary on some machines but
-  # is effectively a no-op when it would not be needed
-  _run sudo snap connect $FRONTEND_NAME:checkbox-runtime $RUNTIME_NAME:checkbox-runtime
-  _run sudo snap connect $FRONTEND_NAME:provider-certification-client $RUNTIME_NAME:provider-certification-client
-  _run sudo snap connect $FRONTEND_NAME:provider-checkbox $RUNTIME_NAME:provider-checkbox
-  _run sudo snap connect $FRONTEND_NAME:provider-resource $RUNTIME_NAME:provider-resource
+    # these are the default plugs (although some custom frontends may not have them)
+    CONNECTIONS="checkbox-runtime provider-checkbox provider-resource provider-certification-client"
```
Hook25, line 90: why don't we use the same line as below to fetch the connections that the runtime support? I'm pretty sure there are more

Hook25 (approved):
LGTM, consider the following to make it a bit easier to read

scriptlets/install_checkbox_snaps
boukeas: Oh, so you mean match the frontend plugs to the runtime slots? (instead of using a predefined set)

boukeas: I have now generalised the way the frontend plugs connect to both runtime slots, as well as other slots available for matching interfaces. No plugs or slots are now hard-coded.

PR #54, opened 2024-11-27
Checkbox-provider-gpgpu is now amd64 only.

pedro-avalos:
Resolves CHECKBOX-1519

checkbox-provider-gpgpu is now `amd64` only. It really only ever worked for `amd64`, this just makes it official.

See: https://github.com/canonical/checkbox/pull/1542


Hook25 (approved):
Ty for changing this, +1

PR #55, opened 2024-12-03
Add: scp scriplets

p-gentili:
This PR adds the `put`/`get` scriplets. Useful in CI when sideloading content or any other necessary asset.

Tested with https://testflinger.canonical.com/jobs/ece65f9a-1240-4444-8a5b-69f0dc85e763

### DIFF
--- a/scriptlets/_get_zapper
+++ b/scriptlets/_get_zapper
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+
+# Transfer files from a remote Zapper via SCP.
+#
+# Description:
+#
+# The IP of the device is provided by the environment variable ZAPPER_IP,
+# which must be set. The user credentials on the remote device is `ubuntu:insecure`
+# by default, unless provided by the environment variables ZAPPER_USER and ZAPPER_PWD.
+#
+# The last argument provided is considered to be the "target", i.e. where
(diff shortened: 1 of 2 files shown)


boukeas:
@p-gentili I am wondering if we should add a `--control` flag to the existing `_get`, `_put` and `_run` scriptlets instead of having separate ones. The amount of code duplication would probably justify that. What do you think?

p-gentili:
> @p-gentili I am wondering if we should add a `--control` flag to the existing `_get`, `_put` and `_run` scriptlets instead of having separate ones. The amount of code duplication would probably justify that. What do you think?

While I agree with you from a code quality perspective, I'd avoid for the moment contaminations in _normal_ scriplets. This is a public repo and I can see us creating something private for host-specific scriplets in the future.

p-gentili:
Hey @boukeas, overnight idea: what if we just update the basic scriplets, adding support for `DEVICE_PWD`? In that case I could just use `DEVICE_IP=$TEST_IP DEVICE_PWD=insecure _run <cmd>`. Same for the other scriplets.

In this way we also remove every host-mention, which is a big plus.

p-gentili:
Closing in favor of https://github.com/canonical/hwcert-jenkins-tools/pull/56

PR #57, opened 2024-12-07
Change: option to skip instaling tools on device

p-gentili:
This PR adds the possibility to not install anything on the DUT, using the `--skip-dut` switch. This is useful when the DUT might be unreachable via SSH, and we plan to interact with it using a support machine.

Tests
1. Default: testflinger.canonical.com/jobs/67d63df5-d0ac-4cfa-96b1-6ade62502d88
2. With `--skip-dut`: testflinger.canonical.com/jobs/95e8639c-72b2-4bd0-8611-e0abe58be865

### DIFF
--- a/install_tools.sh
+++ b/install_tools.sh
@@ -3,7 +3,7 @@
 # Clone the certification tools repo to a local directory.
 # If the repo is already available locally, fetch the latest version.
 # Use the --branch option to specify a specific branch.
-# Use the --skip-dut to skip installing scriptlets to the DUT.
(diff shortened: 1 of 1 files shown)


boukeas (approved):
I think that's generally a useful command-line argument (I know I could have used it in other circumstances) so thanks for adding it.

Just one minor modification I would ask for: I don't think I have used "dut" anywhere in the scriptlets I have written or in their documentation/comments. I usually prefer "device" instead, so I my preference would be `--skip-device` instead of `skip-dut`.

PR #58, opened 2024-12-09
[CERTTF-425] Scriptlets for minimal SRU job

boukeas:
## Description

This PR introduces new scriptlets or amends existing ones in the context of [CERTTF-425](https://warthogs.atlassian.net/browse/CERTTF-425). It is the foundation for its [companion PR](https://github.com/canonical/certification-lab-ci/pull/2), which introduces the workflow using these scriptlets.

### New scriptlets

- `enable_proposed`: enable the proposed repository (for a specific arch and series)
- `install_package_from_proposed`: install a package from a proposed repo, followed by a `dist-upgrade` and a reboot
- `collect_artifacts`: collect Checkbox submission artifacts and (optional) additional files into the `artifacts` folder, to be picked up by Testflinger
- `log`: display timestamped and level-graded messages (for important messages that need to be highlighted in the output or be more easily searchable for), e.g. `2024-12-09 18:20:10 [INFO]: Installing Checkbox debian packages at risk: beta`

### Modifications

- `install_checkbox_deb`: extend so that additional Checkbox providers can be specified
- `install_packages`: extend to support performing a `dist-upgrade` instead of installing specific packages (since much of the `apt-get` command line will be the same)
- `fetch_manifest`: refactor, introduce argument for using a default manifest and rename argument for the output file
- `scenario`: introduce the `--list` argument for retrieving scenario fields with values that are lists
- installer (`install_tools.sh`): simplify to only perform cloning (no fetching) and retrieve the value for `DEVICE_USER` from the scenario, if there is one
- `install_checkbox_snaps`: minor modification to use `log` instead of echo (in line with corresponding changes to `install_checkbox_deb`)

[CERTTF-425]: https://warthogs.atlassian.net/browse/CERTTF-425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scriptlets/scenario
```diff
@@ -24,4 +29,23 @@ dot_to_brackets() {
...
+if [ "$#" -eq 0 ]; then
+    usage
+    echo "Error: missing specifier"
+    exit 1
+fi
+
+RESULT=$(jq -r -e $(dot_to_brackets "$1") "$SCENARIO_FILE")
+EXIT_CODE="$?"
+
+if [ "$EXIT_CODE" -ne 0 ] || [ -z "$ARG_LIST_FLAG" ]; then
```
Hook25, line 46: I can't say I understand what ARG_LIST_FLAG is for

scriptlets/scenario
```diff
@@ -24,4 +29,23 @@ dot_to_brackets() {
...
+    echo "$RESULT"
+else
+    echo "$RESULT" | jq -r 'join(" ")'
```
Hook25, line 49: Is this a footgun waiting to shoot? If we use this in a `$( ... )` then its a space/command injection etc. waiting to happen

scriptlets/sru-helpers/fetch_manifest
```diff
@@ -8,50 +8,100 @@
...
+    """
+    Fetch manifest for `cid` from C3 and write it to the `output` file
+    """
+    # request the manifest from C3
     response = requests.get(
-        f"https://certification.canonical.com/api/v2/machines/{args.cid}/manifest?secure_id={args.secure_id}"
+        "https://certification.canonical.com/api/v2/machines/"
+        f"{cid}/manifest?secure_id={secure_id}"
     )
     if response.status_code != 200:
```
Hook25, line 49: `if not response.ok` (even though I prefer `response.raise_for_status()`

scriptlets/sru-helpers/fetch_manifest
Hook25, line 1: Minor: Overall, sadly python has exceptions, so I believe they are idiomatic, I wouldnt return true false to communicate results

Hook25 (approved):
This lgtm , I have a few requests for more documentation below

scriptlets/collect_artifacts
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+# Find submission files and move them into the "artifacts" directory
+# (this is hard-coded because that's where Testflinger collects them from)
+# Also include any additional files, specified using the --include argument
+
+usage() {
+    echo "Usage: $(basename ${BASH_SOURCE[0]}) [--include <file>*]"
```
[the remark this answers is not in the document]
boukeas: Here's an [example](https://github.com/canonical/certification-lab-ci/blob/f5f799c2cf0df6957a2bc355380ec83231e75db6/jobs/sru/resources/job.generic.yaml#L91).

scriptlets/enable_proposed
```diff
@@ -0,0 +1,44 @@
...
+# enable proposed pocket (on the device)
+log "enabling proposed pocket"
+echo "deb $ARCHIVE_URL $SERIES-proposed main restricted universe multiverse" | _run "sudo tee /etc/apt/sources.list.d/proposed-repositories.list"
+
+# pin package priority (on the device)
+log "pining priority of packages"
+cat << EOF | _run "sudo tee -a /etc/apt/preferences.d/pining"
+Package: '*'
+Pin: release o=Ubuntu,a=$SERIES-proposed
+Pin-priority: 400
```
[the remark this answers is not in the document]
boukeas: No, in order for `install_package_from_proposed` to work the proposed pocket must first be enabled (note the `-t $SERIES-proposed` argument used in `install_package_from_proposed`. I just decided to split the two because (a) I imagine one might want to enable proposed for other purposes and (b) the resulting script would be quite long if it weren't split up.

PR #59, opened 2024-12-20
Change the snap rerun URL

andrejvelichkovski:
After adding the parameter to decide whether to report snap testing jobs to Test Observer or not, we need to update the snap rerunner to include the parameters.

This change modifies the URL used to the alternative with parameters.

### DIFF
--- a/scriptlets/test-executions-rerunner/test_executions_rerunner.py
+++ b/scriptlets/test-executions-rerunner/test_executions_rerunner.py
@@ -81,12 +81,13 @@ class Main:
 
     def _submit_snap_rerun(self, base_job_link: str) -> None:
         rerun_link = f"{base_job_link}/buildWithParameters"
-        logging.info(f"POST {rerun_link}")
-        requests.post(rerun_link, auth=self.jenkins_auth)
+        data = {"TEST_OBSERVER_REPORTING": True}
+        logging.info(f"POST {rerun_link} {data}")
+        requests.post(rerun_link, auth=self.jenkins_auth, json=data)
 
     def _submit_deb_rerun(self, base_job_link: str) -> None:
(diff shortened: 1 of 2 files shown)


boukeas (approved):

Saviq:
My only worry is that if a job does _not_ have the parameter (for whatever reason), the re-run wouldn't work.

Should the rerunner maybe take the parameters from the job it's rerunning, instead?

boukeas:
> My only worry is that if a job does _not_ have the parameter (for whatever reason), the re-run wouldn't work.
> 
> Should the rerunner maybe take the parameters from the job it's rerunning, instead?

@Saviq I can guarantee that triggering from the API without specifying the value for the parameter uses the default value, which is currently True. The only case where this (inexplicably) does not apply is with URL triggering, where the parameter values are unset and this case is handled but the SRU and snap-testing scripts.

Saviq:
@boukeas I'm talking about the case where a job does _not_ have the parameter configured. So the POST would specify a non-existent parameter, which should be considered an error?

But that's probably OK anyway — we can just say that we're expecting the parameter to be there. Using the parameters as-set for the job being re-ran would just be that tiny bit more reliable :)

andrejvelichkovski:
> My only worry is that if a job does _not_ have the parameter (for whatever reason), the re-run wouldn't work.
> 
> Should the rerunner maybe take the parameters from the job it's rerunning, instead?

That's a very good point, I agree the complete solution would be to use the URL based on the job itself.

I still think we should land this so that we unblock anyone using TO for our jobs in the upcoming holidays, and then to keep track of this, I filed [RWT-425](https://warthogs.atlassian.net/browse/RTW-425)

PR #60, opened 2025-01-08
Fix: fallback to checkbox --classic

p-gentili:
Due to `set -e` at the start of the block, the `install_checkbox_snaps` scriplets fails to fallback to checkbox --classic when --devmode is not possible.

This PR temporarily set `set +e` when trying to install the snap in devmode.


boukeas (approved):
I remember not being able to test the fallback to classic because the only device in the snap-testing jobs that uses the classic snap was failing to provision. So this is much appreciated, thank you.

My only minor suggestion is to capitalize the `snap_install_status` variable, for consistency with the rest of the script, which uses this convention for environment variables.

PR #61, opened 2025-01-24
Add support for private checkbox frontends

p-gentili:
This PR adds the capability to install private checkbox frontend snaps. To do so:
1. The REST APIs to snapcraft.io are authorized with a macaroon retrieved from a snapcraft exported login
2. The `snap install` is now `snap download + ack + install`, since `snap download` is the only utility that can benefit from an exported token

---

Resolves [ZAP-1005](https://warthogs.atlassian.net/browse/ZAP-1005)

---

Tested on multiple environments with https://github.com/canonical/hwcert-jenkins-jobs/pull/678

[ZAP-1005]: https://warthogs.atlassian.net/browse/ZAP-1005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cert-tools/get-snapcraft-token.py
Hook25, line 1: this scriplet doesn't get a snapcraft token, it gets a store request authorization header right?

cert-tools/get-snapcraft-token.py
```diff
@@ -0,0 +1,55 @@
...
+    root = Macaroon.deserialize(token_dict["v"]["r"])
+
+    raw_discharge = token_dict["v"]["d"]
+
+    # The discharge macaroon has an expiry: fresh is better
+    refresh_request = Request(
+        REFRESH_ENDPOINT,
+        data={"discharged_macaroon": raw_discharge},
+        method="POST",
+    )
```
Hook25, line 37: I'm not sure what this does, it may be pure ignorance of the whole token mechanism but can you explain why/what we are refreshing here?

scriptlets/install_checkbox_snaps
```diff
@@ -117,17 +124,24 @@ wait_for_snap_changes
...
 set -e
 
 if [ "$SNAP_INSTALL_STATUS" -ne 0 ]; then
     echo "Failed to install $FRONTEND_NAME as a strict snap"
     echo "Installing frontend snap: $FRONTEND_NAME from $FRONTEND_CHANNEL (as a classic snap, using --classic)"
     STRICT_FRONTEND=false
-    _run sudo snap install --no-wait $FRONTEND_NAME --classic --channel=$FRONTEND_CHANNEL
+    _run sudo snap install --no-wait --classic $FRONTEND_NAME.snap
 fi
+_run rm /var/tmp/$TOKEN_FILENAME $FRONTEND_NAME.*|| true
```
Hook25, line 144: Is this enough to purge the token from the machine? also, would it be possible to avoid copying it there?

Hook25 (changes requested):
I have a few questions about this, see if they make sense.

Do we have now a good way to run these jobs on a few "golden" machines we can trust and get back some results @boukeas? In case please do test it out, it looks very good to me

boukeas:
> A first test: http://10.102.156.15:8080/job/pgentili-rpi4b8g-arm64-core24-beta/13/console

Are you sure this log is indicative of a successful test?
```
13:24:17 8    Doing   today at 13:23 UTC  -                   Install "checkbox" snap from file "checkbox.snap"
13:24:17 retry: 'check_for_snap_changes' returned 1, backing off for 30 seconds and trying again...
13:24:51 No ongoing or pending snap changes
13:24:51 error: snap "checkbox" not found
13:24:51 ++ _run snap list checkbox
13:24:51 ++ tail -n1
13:24:51 ++ awk '{print $3}'
13:24:51 error: no matching snaps installed
13:24:51 + CHECKBOX_REV=
```

boukeas:
> Do we have now a good way to run these jobs on a few "golden" machines we can trust and get back some results @boukeas? In case please do test it out, it looks very good to me

I don't know if I am interpreting the question correctly but I don't think there is a single set of jobs/machines that would be appropriate for testing every change, because different changes would affect different jobs. A change with a wide impact (such as the one introducing `install_checkbox_snaps` to snap-testing jobs required a [very wide range of test jobs](https://github.com/canonical/hwcert-jenkins-jobs/pull/618)). In this case, we would need to test jobs where the checkbox snap is `checkbox-` but also cases where it isn't.

scriptlets/install_checkbox_snaps
```diff
@@ -117,17 +124,24 @@ wait_for_snap_changes
 echo "Installing frontend snap: $FRONTEND_NAME from $FRONTEND_CHANNEL (as a strict snap, using --devmode)"
 STRICT_FRONTEND=true
 
+if [[ -n "$TOKEN_FILENAME" ]]; then
+    echo "A snapcraft token will be used to download the $FRONTEND_NAME snap"
+    _put $UBUNTU_STORE_AUTH_DATA_FILENAME /var/tmp
+fi
+_run ${TOKEN_FILENAME:+ UBUNTU_STORE_AUTH_DATA_FILENAME=/var/tmp/$TOKEN_FILENAME} snap download $FRONTEND_NAME --channel=$FRONTEND_CHANNEL --basename=$FRONTEND_NAME
```
boukeas, line 131: My main question is around this point: Is there an advantage to _always_ downloading the snap, regardless of its origin (private store or not), as opposed to _only_ downloading when the snap comes from a private store?

scriptlets/install_checkbox_snaps
boukeas: > would it be possible to avoid copying it there?

@p-gentili Can't we use something like [this](https://snapcraft.io/docs/snapcraft-authentication#snapcraft_store_credentials-environment-variable)?

scriptlets/install_checkbox_snaps
```diff
@@ -28,7 +30,10 @@ get_runtime() {
     STORE=$(_run "snap model --assertion" | sed -n 's/^store:\s\(.*\)$/\1/p')
     # get the snap info from the snap store and extract the base
     SNAPSTORE_URL="https://api.snapcraft.io/v2/snaps/info/$FRONTEND_NAME?fields=base"
-    SNAPINFO=$(curl -s -H 'Snap-Device-Series: 16' ${STORE:+ -H "Snap-Device-Store: $STORE"} "$SNAPSTORE_URL")
+    SNAPINFO=$(curl -s -H 'Snap-Device-Series: 16' \
+        ${STORE:+ -H "Snap-Device-Store: $STORE"} \
+        ${TOKEN_FILENAME:+ -H "Authorization: $($TOOLS_PATH/cert-tools/get-snapcraft-token.py $UBUNTU_STORE_AUTH_DATA_FILENAME)"} \
```
boukeas, line 35: I know it would make this even longer, but I think I would prefer adding some more structure to the way these `cert-tools` are organized (which is unfortunately badly missing in other parts of the repo). So, I would suggest moving the `get-snapcraft-token.py` under `cert-tools/snapcraft` or something like that.

boukeas (changes requested):

scriptlets/install_checkbox_snaps
p-gentili: The token is required on the host running `snapcraft`/`snap download`. My first idea was to download the snap on the agent and copy it to the machine but, AFAIK, installing snapcraft/snap is not an option.

p-gentili: I moved it right after `snap download`, to reduce as much as possible the possible exit failures in between.

cert-tools/get-snapcraft-token.py
p-gentili: yes, the full documentation is linked on the module-level docstring but I added something there as well

cert-tools/get-snapcraft-token.py
```diff
@@ -0,0 +1,55 @@
...
+REFRESH_ENDPOINT = "https://api.snapcraft.io/v2/tokens/refresh"
+
+
+def get_authorization_header(token: str) -> str:
+    """Return the required authorization header."""
+
+    token_str = base64.b64decode(token).decode("utf-8")
+    token_dict = json.loads(token_str)
+
+    if not token_dict["v"] and token_dict["v"]["r"] and token_dict["v"]["d"]:
```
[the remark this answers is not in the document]
p-gentili: I couldn't find any schema, but I added a comment on what I've found looking at snapd/craft source codes.

p-gentili:
> Calling that API in an authenticated manner might be trickier than expected...

Finally managed to refresh the token.

Another test: http://10.102.156.15:8080/job/pgentili-rpi4b8g-arm64-core24-beta/15

> Are you sure this log is indicative of a successful test?

I've debugged this and it's a problem with the snap itself, it fails at the `install` hook. I'll retry as soon as I fix the snap.

scriptlets/install_checkbox_snaps
p-gentili: > @p-gentili Can't we use something like [this](https://snapcraft.io/docs/snapcraft-authentication#snapcraft_store_credentials-environment-variable)?

To clarify further: `snap download` expects the file path, not the content.

boukeas: I have really struggled to locate any corresponding documentation but I did find [this](https://forum.snapcraft.io/t/improvements-in-snap-download/1422/13), which indicates you can also use `UBUNTU_STORE_AUTH` with the contents, instead of the file. Do you think it's worth a try? It might save us moving the file around and deleting it.

boukeas:
> EDIT: this is a test with a public snap http://10.102.156.15:8080/job/pgentili-rpi4b8g-arm64-core24-beta/16/

What I often do is compare the results with the [corresponding one in the main branch](http://10.102.156.15:8080/job/cert-rpi4b8g-arm64-core24-beta/lastSuccessfulBuild): there are some additional tests failing in the run from the feature branch and we need to make sure these are not due to the changes in this PR. For example, in this particular case, the additional failures are GPIO-pin related and this might be connected to the fact that the `install_checkbox_snaps` script tries to connect plugs and sockets in the best way it can but it does stop connecting after the first match in each category and this may affect GPIO pins. This might be OK after running the post-install hook but it might cause issues when using `snap download` since the post-install hook is not ran.

scriptlets/install_checkbox_snaps
```diff
@@ -104,6 +109,8 @@ log "Installing Checkbox snaps"
 export RISK=$1
 export FRONTEND_NAME=$2
 export FRONTEND_TRACK=$3
+export TOKEN_FILENAME=$(basename $UBUNTU_STORE_AUTH_DATA_FILENAME 2> /dev/null)
+
```
boukeas, line 113: This is very minor but it messes with my pedantic nature: I would rather this line setting `TOKEN_FILENAME` were moved after the line that sets `RUNTIME_CHANNEL`. My reasoning is that the first three exports start with the primary inputs and the following lines deduce a set of values that are directly relevant to them. I see them as a block. `TOKEN_FILENAME` is not relevant so I would rather move it after the block.

boukeas (changes requested):
I am very skeptical about downloading instead of installing even when we don't really need to.

I understand it unifies the approach and keeps the code simpler but it means that we are replacing code (i.e. snap-downloading instead of snap-installing) that has been thoroughly tested in practice and now needs to be re-tested. As I understand it, snap-downloading means the post-install hooks are not ran and I don't know if this might have adverse effects.

So I would personally feel safe to approve if this PR included a range of snap-testing jobs that don't use the private checkbox frontend, including many of the jobs that use frontends from brand stores. From my experience when I was testing [this PR](https://github.com/canonical/hwcert-jenkins-jobs/pull/618), many of the brand-store jobs present their own unique challenges.

scriptlets/install_checkbox_snaps
p-gentili: I agree but `TOKEN_FILENAME` is used by `get_runtime`...

scriptlets/install_checkbox_snaps
p-gentili: The documentation is a jungle. But what you suggest works, and also solves another point you raised below. Thanks!

scriptlets/install_checkbox_snaps
p-gentili: Problem solved thanks to `UBUNTU_STORE_AUTH`

p-gentili:
> I understand it unifies the approach and keeps the code simpler but it means that we are replacing code (i.e. snap-downloading instead of snap-installing) that has been thoroughly tested in practice and now needs to be re-tested. As I understand it, snap-downloading means the post-install hooks are not ran and I don't know if this might have adverse effects.

I get you, this is a big change and there's a risk we break our test runs. But I also think that making the script more complicated (it would have 4 branches!) doesn't really work in the long run. You should also consider this particular frontend will be THE one we use in SRU jobs.

> So I would personally feel safe to approve if this PR included a range of snap-testing jobs that don't use the private checkbox frontend, including many of the jobs that use frontends from brand stores. From my experience when I was testing [this PR](https://github.com/canonical/hwcert-jenkins-jobs/pull/618), many of the brand-store jobs present their own unique challenges.

This is very needed, I agree, and I'll report the results as soon as I get back from vacation.

boukeas:
> I've [tested many environments](https://github.com/canonical/hwcert-jenkins-jobs/pull/678), like you @boukeas did last time.

That's brilliant! I thinks it's important that you add this link to the description of the PR (otherwise future generations might form the impression that testing was limited). Besides, that's a companion PR, it needs to be linked.

Also, I'm making a note that we need to approve this PR _before_ approving the companion because then the companion will require a final commit to install from `main` rather than this branch.

boukeas (changes requested):
I think this has ended up looking very good and I feel much more confident about it since it has been tested so extensively. I have left a couple of minor comments on... comments.

I wanted to ask if you have ran any test on this for the `--classic` branch? My specific concern is that we only run `connect` in the "strict" `--devmode` case. I am wondering if it now needs to be ran in every case.

scriptlets/install_checkbox_snaps
```diff
@@ -24,11 +25,12 @@ get_ubuntu_arch() {
 }
 
 get_runtime() {
-    # get the store token from the device, if available
```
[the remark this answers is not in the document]
p-gentili: Done in https://github.com/canonical/hwcert-jenkins-tools/pull/61/commits/143fb7c92fc91f58ed9f78afb97c76b6b3b5e1fe

scriptlets/install_checkbox_snaps
```diff
@@ -117,16 +120,20 @@ wait_for_snap_changes
 echo "Installing frontend snap: $FRONTEND_NAME from $FRONTEND_CHANNEL (as a strict snap, using --devmode)"
 STRICT_FRONTEND=true
 
+_run ${STORE:+ UBUNTU_STORE_ID=$STORE} \
+    ${UBUNTU_STORE_AUTH:+ UBUNTU_STORE_AUTH=$UBUNTU_STORE_AUTH} \
+    snap download $FRONTEND_NAME --channel=$FRONTEND_CHANNEL --basename=$FRONTEND_NAME
+_run sudo snap ack $FRONTEND_NAME.assert
```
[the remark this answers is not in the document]
p-gentili: Done in https://github.com/canonical/hwcert-jenkins-tools/pull/61/commits/585658413a89f6575c3a175f94202283ac905b0a

boukeas (approved):

Hook25 (approved):

PR #64, opened 2025-02-25
(DRAFT) Change: update only mainboard

p-gentili:
Updating add-ons every time seems to be an overkill: we rarely update it w/o bringing new functionalities to existing add-ons. I think for now that artifact shouldn't be part of the validation workflow.


p-gentili:
> Is this pending the firmware hashing process we disussed in the hague?

yes, right now versioning is still a manual operation but that's the way we're headed to

PR #68, opened 2025-03-13
Fix test executions rerunner

boukeas:
## Description

Rerun requests for charms crashed the test execution rerunner because they contained `ci_link` entries with a `None` value. The existing tests for the rerunner covered the case where `ci_link` was missing but not the case where it was `None`. This PR addresses the issue.

Following the test process described below, it was also discovered that certain URLs contained an unexpected plus character (`+`) that resulted in the URLs not being matched by the regular expressions being used. This issue is also addressed.

## Tests

The offending rerun requests were retrieved from the logs and parsed locally, following the process in the README. The problem was reproduced and then resolved. The `test-executions-rerunner` Jenkins jobs was modified to retrieve the script from this branch and the rerun requests started going through.


omar-selo (approved):
Great work catching this!

PR #69, opened 2025-04-08
chore: handle the case where `get_runtime` cannot determine runtime

boukeas:
## Description

The `install_checkbox_snaps` scriptlet contains a `get_runtime` function for determining the Checkbox runtime. The current code for `get_runtime` has two issues:

- The code essentially assumes that the Checkbox runtime can always be determined, which is unfortunately not the case. When the Checkbox runtime, in the form `checkboxYY`, cannot be determined, the absence of a `YY` suffix results in `checkbox` being installed instead of an error being raised.
- The Checkbox runtime is determined from the `base` property of the Checkbox frontend. This, in turn, is retrieved through the `info` endpoint of the snap Store API, which will _not_ return results when the requested channel does not contain revisions and is instead tracking a more stable channel (see the Tests section for an example).

This PR:
- modifies the code for `get_runtime` so that an error is raised in the case where the Checkbox runtime cannot be determined
- introduces a Python tool for retrieving snap information from the `refresh` endpoint, which overcomes the limitations of the `info` endpoint
- modifies the code for `get_runtime` so that this Python tool is used instead of a `curl` to the `info` endpoint.

### Tests for the `snapstore` package

```
$ snapstore-info checkbox uc20/beta amd64 --fields base version
{"base": "core20", "version": "4.4.0-dev94", "effective-channel": "uc20/beta"}
```
Here is an example where `checkbox-mir` has no revisions on `latest/candidate`. Using the `info` endpoint will fail:
```
$ snapstore-info checkbox-mir latest/candidate arm64 --fields base --use-info
Traceback (most recent call last):
  File "/home/boukeas/.local/bin/snapstore-info", line 8, in <module>
    sys.exit(cli())
  File "/home/boukeas/.local/pipx/venvs/snapstore/lib/python3.10/site-packages/snapstore/info.py", line 167, in cli
    result = retriever(
  File "/home/boukeas/.local/pipx/venvs/snapstore/lib/python3.10/site-packages/snapstore/info.py", line 115, in info
    raise ValueError(
ValueError: No info for checkbox-mir=latest/candidate on arm64
```
Using the `refresh` endpoint (the default) succeeds:
```
$ snapstore-info checkbox-mir latest/candidate arm64 --fields base 
{"base": "core22", "effective-channel": "latest/stable"}
```
The tool can handle snaps in brand stores, or snaps published in branches:
```
$ snapstore-info hbt-imx-kernel latest/beta/stream2 arm64 --fields base revision version --store ged8Ce2jo8nohkieshah
{"base": null, "revision": 385, "version": "5.4.0-1068.80.1", "effective-channel": "latest/beta"}
```
It can also handle private snaps:
```
UBUNTU_STORE_AUTH=... snapstore-info checkbox-extra latest/beta arm64
{"created-at": "2025-03-28T11:18:05.164588+00:00", "download": {"deltas": [], "sha3-384": "920a821828633671bedce5943b011b623ebf26fda6950c885ab65a917625edbf30b5ef6bcc1e407e9a7b43e1cb55deb2", "size": 24576, "url": "https://api.snapcraft.io/api/v1/snaps/download/lC1pezzk2DmnVMOyiuWavnr5P3E51KTk_385.snap"}, "license": "unset", "name": "checkbox-extra", "prices": {}, "publisher": {"display-name": "Canonical Certification Team", "id": "Euf8YO6waprpTXVrREuDw8ODHNIACTwi", "username": "ce-certification-qa", "validation": "unproven"}, "revision": 385, "snap-id": "lC1pezzk2DmnVMOyiuWavnr5P3E51KTk", "summary": "Checkbox tests for extra", "title": "checkbox-extra", "type": "app", "version": "0.0.1", "effective-channel": "latest/beta"}
```

### Test for `install_checkbox_snaps`

This is part of a log from a [Jenkins job](http://10.102.156.15:8080/job/cert-zhongyi-snapd-beta/54/console) where `install_checkbox_snaps` failed to identify the Checkbox runtime.

```
07:52:40 + install_checkbox_snaps checkbox-zhongyi=latest/beta
07:52:40 2025-04-11 06:52:34 [INFO]: Installing Checkbox snaps
07:52:40 Installing runtime snap: checkbox from latest/beta
```
The reason is that there is no revision published for `checkbox-zhongyi` on `latest/beta` (the "effective-channel" is `latest/stable`) so the `info` endpoint returns no relevant entry. When the job is reconfigured to install the tools from this branch instead of `main`, the [log](http://10.102.156.15:8080/job/cert-zhongyi-snapd-beta/77/consoleFull) shows that the issue is resolved:
```
21:16:37 + install_checkbox_snaps checkbox-zhongyi=latest/beta
21:17:11 2025-04-12 20:16:26 [INFO]: Installing Checkbox snaps
21:17:11 Installing runtime snap: checkbox20 from latest/beta
```


scriptlets/install_checkbox_snaps
```diff
@@ -60,33 +60,24 @@ get_runtime() {
...
     else
       # base coreYY -> runtime checkboxYY
-      # (the final "grep ." will cause the command to fail if there's no match)
-      RUNTIME_NAME="checkbox$(echo $BASE | sed -n 's/^core\([0-9]\{2\}\)$/\1/p' | grep .)"
+      RUNTIME_SUFFIX=$(echo $BASE | sed -n 's/^core\([0-9]\{2\}\)$/\1/p')
     fi
-    echo $RUNTIME_NAME
+    if [ -z "$RUNTIME_SUFFIX" ]; then
+        echo "Unable to determine runtime for $FRONTEND_NAME=$FRONTEND_TRACK/$RISK"
+        return 1
```
p-gentili, line 78: Thank you!

cert-tools/snapstore/src/snapstore/info.py
```diff
@@ -0,0 +1,178 @@
...
+            params=params,
+            store=store,
+        )
+
+    def raw_refresh_one(
+        self,
+        snap: str,
+        channel: str,
+        architecture: str,
+        store: Optional[str] = None,
```
p-gentili, line 53: I've noticed that when there is a dedicated snap store, and the checkbox frontend snap is not included in the inclusion list, the API will fail. At least, `info` does.

We could maybe retry w/o `store` in case of failure? 

If we go that way though, the following `snap download` should have the `SNAP_STORE` variable unset... so it's a bit trickier than a simple retry.

scriptlets/install_checkbox_snaps
p-gentili, line 1: This is much easier to read now, thank you!

p-gentili (commented):
Thank you very much for taking care of this George - there are so many combinations of snaps and stores and this make is much easier to read and debug.

I haven't reviewed the auth/client modules because I think you could consider using [craft-store](https://github.com/canonical/craft-store) which we'll makes us always up-to-date with the store / ubuntu-one changes. TBH, I've not been able to make it work last time I tried, but what you're proposing is not a *simple* script anymore, so it might be worth it to give it another try.

And finally, feel free to get rid of `get-auth.py`.

cert-tools/snapstore/src/snapstore/client.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        if headers is None:
+            headers = {}
+        authorization = (
+            AuthClient.authorization_from_token(self.token)
+            if self.token else None
+        )
+        return {
+            "Snap-Device-Series": "16",
+            **headers,
```
p-gentili, line 41:
```suggestion
        authorization = (
            AuthClient.authorization_from_token(self.token)
            if self.token else None
        )
        return {
            "Snap-Device-Series": "16",
            **(headers if headers else {}),
```
just to follow the same style as below

p-gentili (changes requested):
If we ever want to switch to another authentication method, let's not forget about the `craft-store` package. 

Everything looks good and clear to read. I left a comment below, feel free to get not consider it if you don't like it, but I would remove the `get-auth.py` script before landing.

cert-tools/snapstore/src/snapstore/client.py
boukeas: Done.

cert-tools/snapstore/src/snapstore/info.py
p-gentili: We agreed to follow up on this.

p-gentili (approved):
LGTM!

PR #71, opened 2025-04-23
Cache version published intermediate results

Hook25:
When iterating over channels/binary build outputs of the same package, this script re-requests the status every time. It makes sense to eventually invalidate the data and re-load it, but to do so every time is very wasteful (and slow).

This implements a cache for both fetching so that:
- checks on the same snap (but different channel) are cached
- checks on the same package (but different distribution/arch) are cached

Given that the script ping pongs from checking snaps to checking debs, the cache is invalidated once every cycle.

Minor: This updates the ubuntu version in the github workflow (rip ubuntu20)

### DIFF
--- a/.github/workflows/test_version_published.yaml
+++ b/.github/workflows/test_version_published.yaml
@@ -16,7 +16,7 @@ jobs:
     defaults:
       run:
         working-directory: version-published
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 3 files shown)


fernando79513 (commented):
Good job!
We will reduce significantly the number of calls this way, specially in the case of the packages.
LGTM +1!

PR #72, opened 2025-05-01
[RTW-514] Update README with an overview of the essential scriptlets

boukeas:
Does what it says on the tin.


Hook25 (approved):
LGTM, it really does do what it says on the tin

PR #73, opened 2025-05-02
mailtool: add Date field

pinaraf:
Our mail servers don't always add the Date field, so instead force it when generating the email.


rmartin013 (approved):
Thanks for the change. @canonical/cert-orchestration-reporting @p-gentili Pierre initiated this change because he has troubles with his mail client. It turns out the tool don't send mail with proper date metadata, which looks like (to me) something we should fix.

PR #79, opened 2025-07-08
Don't put submission.json into artifact after testing.

kevinyehk:
I noticed that the size of `submission.json` has grown significantly, which consumes too much disk space on our jenkins server.
Since we already have the submission.tar.xz in the artifact, I don't think is necessary to extract the json from the tar file.

This PR addresses https://warthogs.atlassian.net/browse/LM-1941.


Hook25 (approved):
+1

PR #105, opened 2026-04-09
new: allow to skip the checkbox controller installation

p-gentili:
Using toolbox is pretty useful when dealing with a freshly provisioned device. 

When using the install-checkbox-* scripts, the agents gets a version of checkbox-cli using `pipx`. I'd like to avoid that when running the script from my machine, and just run checkbox from source. 

In practice, added `--skip-agent-install` to install on machine, and return.


Hook25 (changes requested):
Note, this doesn't skip installing the agent, it skips installing the controller on the (TF) agent (sigh). Change it to `--skip-controller-install`

p-gentili:
sorry, I work too much on Testflinger these days (that's the agent I meant)

Hook25 (approved):

PR #115, opened 2026-07-02
Ensure stable snaps

Hook25:
# Description

This scriplet ensures that all snaps that are pre-installed in the image are refreshed to the latest available stable revision in the stable channel they are tracking. 

For example, snapd latest/stable will be refreshed to latest/stable (simulating an instant autorefresh to the tracked channel), while bluez 22/stable will be refreshed to 22/stable.

This also "downgrades" to stable any snap that is not stable (without changing the track as well), this cleans up the setup from any previous manual modification that may have been made.

# Testing
```
lxc launch ubuntu:resolute resolute
# set your DEVICE_IP/DEVICE_USER using lxc ls
lxc exec resolute -- snap install hello-world
lxc exec resolute -- snap install checkbox --channel=24.04/beta
ensure-stable-snaps
```

### DIFF
--- a/cert-tools/toolbox/src/toolbox/interfaces/snaps.py
+++ b/cert-tools/toolbox/src/toolbox/interfaces/snaps.py
@@ -10,7 +10,6 @@ from toolbox.interfaces.status import SystemStatusInterface
 from toolbox.results import BooleanResult
 from toolbox.retries import retry, RetryPolicy
 
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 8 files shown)


cert-tools/toolbox/src/toolbox/interfaces/snaps.py
```diff
@@ -1,6 +1,7 @@
 """Interface for managing snap packages on devices."""
 
 from functools import partial
+from typing import Optional, List
```
p-gentili: Since this is 3.10+, you can use `list` and `| None` instead. 

EDIT, ok I've noticed you prefer `Optional`, fine. I would use `list` though instead.

cert-tools/toolbox/tests/cli/test_ensure_stable_snaps.py
p-gentili, line 1: Can we add an example with `branch` defined?

p-gentili (commented):
Very handy, thanks!

p-gentili (approved):

